PR #16, opened 2022-03-03
Small fix (line 25)

kant:


hlorus:
Great, thanks for comtributing!

PR #18, opened 2022-03-06
Minor fix (paragraph 26)

kant:


hlorus:
Note that the building process of the documentation isn't yet automated so i will have to do it manually before the changes will be visible.

PR #32, opened 2022-04-24
Fix cache py3.8

marclava:
At install Python complained about the cache decorator.
As documented here: https://docs.python.org/3/library/functools.html
the cache decorator "Returns the same as lru_cache(maxsize=None)"
So I fixed the issue with a conditional import.


hlorus:
Awesome, thanks alot for contributing!

PR #130, opened 2022-05-16
add support for Diameters leader lines insode or outside a radius; make Diameter dimension value mobile

exegetor:


class_defines.py
```diff
@@ -1,3 +1,4 @@
+from distutils import core
```
hlorus, line 1: This looks like a leftover import

hlorus (commented):
Generally this looks fine. The draw_inside properties of the distance and angle constraints are probably WIP. The patch could already be merged if they are removed from it, however if you prefer you can also just update whenever those changes are done as well.

class_defines.py
exegetor: good observations, thanks for the review.   PR updated to fix.

hlorus:
Alright this is already a great improvement and can go in like that. I'm just pointing a few things out which will be great additions but can come in later:
- Improving the default placement position to be on the outside
- Changing the operator on the text to the same as the line so it can be dragged from the text as well.

Issue #131, opened 2022-05-16
Align View with sketch [REQUEST]


mikemac30:
**Motivation / Existing Problem**
All cad programs I've used in the past automatically align the viewport camera to the current sketch. Usually you can still move the view around but by default they typically aligns the view to be normal to the sketching plane. This makes editing the sketch easier because the sketch isn't warped by perspective.

**Description**
After clicking "Add Sketch", or the edit icon "Set Active Sketch", the blender viewport should align itself to be normal to the sketching plane.

hlorus:
Id suggest to have a focus-sketch operator and a preference to auto execute that on sketch enter.

hlorus:
This is already implemented but hidden under an experimental setting. To activate it set the "Show Debug Settings" in the addon preferences, then activate the "Align View" setting.

There are still some visual glitches during or after the transition, at least on my setup. The code can be found here:
https://github.com/hlorus/CAD_Sketcher/blob/main/operators/utilities.py#L85

It would probably also make sense to put the functionality into it's own operator.

kaiwas:
I'm looking forward to the release of SketcherCAD. With this tool, Blender's capabilities will greatly increase.

Auto-alignment is a very useful thing, but I hope that there will always be an option to turn it off.

Off-topic question - maybe you know how to turn off auto-alignment in FreeCAD? (I think you are involved in the development)
It's just awful that there is no way to turn this option off.

hlorus:
> @hlorus  Off-topic question - maybe you know how to turn off auto-alignment in FreeCAD? (I think you are involved in the development)
> It's just awful that there is no way to turn this option off.

No clue if it's possible to turn that off. I'm involved with the development of the measure facility only so i cannot really help with that, sorry.

kaiwas:
> i cannot really help with that, sorry.

The measure facility out great!
As for the question - nothing to worry about. No need to apologize ;).
Thanks for the answer! I wish you success!

PR #132, opened 2022-05-16
Align Viewport to Sketch

mikemac30:
Pull request to add this feature: #131 

This would only align the view when new sketches are created, however this could easily be adopted to editing sketches as well.
CADSketcher

### DIFF
--- a/operators.py
+++ b/operators.py
@@ -12,7 +12,7 @@ from bpy.props import (
     BoolProperty,
 )
 import math
-from mathutils import Vector, Matrix
+from mathutils import Vector, Matrix, Euler
 from mathutils.geometry import intersect_line_plane, distance_point_to_plane
 
 import functools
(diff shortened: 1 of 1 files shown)


hlorus:
Hey there, nice to see someone is working on this, definitively a great addition. I gave this a quick go and it seems to work great, at least on the base workplanes. There is something going wrong when using it on workplanes that are not placed on the origin however.
  
view_to_wp_bug
 
Also note that you can simply access the active region_3d by context.region_data instead of iterating through all areas, unless this is by design...

mikemac30:
Thanks yeah, just rewrote it a bit. It looks like its working for workplanes with random rotations as well now. Also since this version also tilts the camera to align with the wp, it resets the view after leaving the sketch.

operators.py
```diff
@@ -2726,12 +2734,26 @@ def activate_sketch(context, index, operator):
...
+        context.region_data.view_rotation = quat
+        context.region_data.view_location = sk.wp.p1.location #lookat this point
+        context.region_data.view_distance = 6 # from this far away
+
     else:
         space_data.show_object_viewport_curve = True
         space_data.show_object_viewport_mesh = True
+        
+        #Reset view
+        context.region_data.view_rotation = (Euler((0.7854,0,0.7854*3),'XYZ').to_quaternion())
```
hlorus, line 2754: Generally it's better to avoid using magic numbers and give them a symbolic name. In this case it could be something like this: PI_DIV_4 = math.pi / 4

hlorus (commented):
It looks like the view isn't always properly oriented, when testing the X-axis of a sketch goes from bottom to top, IMO it should go from left to right and the Y-axis should go from bottom to top.

Note that it might not even be necessary to do calculations here. workplanes already have a matrix_basis method which is used everywhere. It should probably be possible to use it together with an offset matrix or the view_distance property, e.g.:
rv3d.view_matrix = entity.matrix_basis.inverted() + offset_mat

hlorus:
As this is now planned for 1.0 i've commited this with some edits, see: a41f555b8f682758e6385d131c72748bab99cf8c.

PR #134, opened 2022-05-20
enable dragging dimension constraints by text

exegetor:
It's hard to believe it is this simple, but it tests out well and doesn't seem to interfere with anything else.
This sure does improve usability... like, lots!


hlorus:
It is actually that simple, thanks for the PR.

PR #156, opened 2022-05-30
Radius dimensioning for arcs

exegetor:


hlorus (commented):
I didn't yet give this an actual try but here are some first thoughts:
I'd say the radius mode should be implemented in the actual constraint so that if it's in radius mode the constraint.value is the radius and the solvesystem will recive the value multiplied by two as there's only a diameter constraint.

The constraint mode should be exposed to the user, in the context menu of the constraint and ideally in the diameter operator settings.

Regarding the 2 part gizmo shape: take a look at the [docs](https://docs.blender.org/api/current/bpy.types.Gizmo.html#bpy.types.Gizmo.new_custom_shape). This should be trivial if the lines type is used for new_custom_shape as you're simply defining lines by point pairs.

hlorus:
Alright, thanks for your work, i've made some changes and merged the PR.

Issue #53, opened 2022-04-27
Allow entity naming


hlorus:
Add functionality to name created entities similar to sketches. The name should probably be editable from the entity's context menu and from the entity browser.

hlorus:
This has been implement a while ago.

PR #175, opened 2022-06-10
Support entity naming

BaoNTGH:
- By default, name will be generated by combining type and index.
- Naming could be done through entity's context menu and also in entity
  browser
- In entity browser, selector had been changed to the radio button in first column
- name is introduced as StringProperty on SlvsGenericEntity as a wrapper of name property to provide default name value for entities saved by previous CAD Sketch plugin where name were not populated

This is to address issue #53 

Change name in context menu
image

Change name in entity browser
image

Selection mode
image

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -27,8 +27,15 @@ from .functions import unique_attribute_setter
 logger = logging.getLogger(__name__)
 
 
+def entity_name_getter(self):
+    return self.get("name", str(self))
+
+def entity_name_setter(self, new_name):
+    self["name"] = new_name
+
 class SlvsGenericEntity:
     slvs_index: IntProperty(name="Global Index", default=-1)
+    name: StringProperty(name="Name", get=entity_name_getter, set=entity_name_setter, options={"SKIP_SAVE"})
     fixed: BoolProperty(name="Fixed")
     visible: BoolProperty(name="Visible", default=True, update=functions.update_cb)
     origin: BoolProperty(name="Origin")
@@ -1164,7 +1171,6 @@ class SlvsCircle(SlvsGenericEntity, PropertyGroup, Entity2D):
         nm (SlvsNormal2D):
         sketch (SlvsSketch): The sketch this entity belongs to
     """
-
     radius: FloatProperty(
         name="Radius",
         description="The radius of the circle",
@@ -1278,7 +1284,6 @@ class SlvsCircle(SlvsGenericEntity, PropertyGroup, Entity2D):
 slvs_entity_pointer(SlvsCircle, "nm")
 slvs_entity_pointer(SlvsCircle, "ct")
 slvs_entity_pointer(SlvsCircle, "sketch")
-
 
 def update_pointers(scene, index_old, index_new):
     """Replaces all references to an entity index with it's new index"""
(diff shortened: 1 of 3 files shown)


hlorus:
Awesome! Regarding the entity_name prop, did you try using "name" as property and id did not work? I'd say it shouldn't be a problem at all, it also seems that blender always adds a name property.

If there are problems with previous versions of the addon versioning code can be used to fix it.

BaoNTGH:
> Awesome! Regarding the entity_name prop, did you try using "name" as property and id did not work? I'd say it shouldn't be a problem at all, it also seems that blender always adds a name property.
> 
> 
> 
> If there are problems with previous versions of the addon versioning code can be used to fix it.

Yeah, as name is currently used in SlvsSketch and it also inherited from SlvsGenericEntity. If I introduce name property, Sketch name got overwritten. I might e oversighting something.
Let me know if you have any pointer

hlorus:
Yeah this makes actually sense because the sketch property group class currently inherits the name property from the propertygroup type probably. So there will be some multi-inhertitance confusion.
Anyway that should be fixed by explicitly defining the name prop (without the custom getter/setter) on the sketch class as well.

BaoNTGH:
> Yeah this makes actually sense because the sketch property group class currently inherits the name property from the propertygroup type probably. So there will be some multi-inhertitance confusion. Anyway that should be fixed by explicitly defining the name prop (without the custom getter/setter) on the sketch class as well.

Sketch is using unique_attribute_setter which is a whole another thing. That unique_attribute_setter doesn't work well with other entities due to the way it set attributes. Currently the settter/getter on Generic Entity already handled that so I would want to avoid changing the Sketch for now. The unique attribute setter could be retired in future I believe.

hlorus:
> > Yeah this makes actually sense because the sketch property group class currently inherits the name property from the propertygroup type probably. So there will be some multi-inhertitance confusion. Anyway that should be fixed by explicitly defining the name prop (without the custom getter/setter) on the sketch class as well.
> 
> Sketch is using unique_attribute_setter which is a whole another thing. That unique_attribute_setter doesn't work well with other entities due to the way it set attributes. Currently the settter/getter on Generic Entity already handled that so I would want to avoid changing the Sketch for now. The unique attribute setter could be retired in future I believe.

I missed that comment, sorry. The unique attribute setter is just a fancy way to ensure that specified attributes are unique across a collection. So for sketches we want a unique name so that it can later be used to name objects. Not quite sure why you would want to get rid of it.

hlorus:
Merged with minor ui tweak 584c10d50961852f1886ce54ce1967eaf01f2c33 2b6d513955f53980bbaf6693685d0ebdfe8c3705

BaoNTGH:
> I missed that comment, sorry. The unique attribute setter is just a fancy way to ensure that specified attributes are unique across a collection. So for sketches we want a unique name so that it can later be used to name objects. Not quite sure why you would want to get rid of it.

Yup, that's what I understood (I could be wrong) about the purpose of unique_attribute_setter. However, if we consolidate the logic to set unique name, we might not need that method anymore and use the same way as we generate the name for the entities (not necessarily type + index, can be anything). unique_attribute_setter iterfere into how property is set, for example:

entity.param = "something" => this will trigger unique_attribute_setter where it tries to set entity['param'] = "something". Anywhere else down the line try to access entity.param, it yields the error of "param is not a property on entity". This is not intuitive for dev and it removes some kinds of "strong type" and compliation error which is harder for development. Just my 2 cents.

Issue #58, opened 2022-04-27
Entity conversion resolution 


hlorus:
Right now converted meshes approximate curves with a given resolution. Ideally this resolution could be configured as users might want more detail on the converted mesh.

Since mesh converting is currently based on the bezier converter this simply boils down to setting the correct values in the bezier datablock before the to_mesh()  method is called. Note that setting this property by the user doesn't work because the bezier object is temporary an gets recreated every time conversion happens.

This setting could probably be put in one of a few different scopes e.g. one setting for the addon/ per sketch/ per spline. For now it should be sufficient to simply have it per sketch.

SupinePandora43:
What about `multires` support too?

Cheaterman:
Hello! Sorry if this should be obvious, but I don't really understand what this feature is about ; could it be possible to add some additional information, given this is tagged as "good first issue"? I'm thinking "current behavior", "expected behavior", and some pointers as to where in the codebase the change should be made?

It would be awesome if such extra information could be added to other issues with the "good first issue" tag too, please :-) thanks in advance!

SupinePandora43:
> Hello! Sorry if this should be obvious, but I don't really understand what this feature is about ; could it be possible to add some additional information, given this is tagged as "good first issue"? I'm thinking "current behavior", "expected behavior", and some pointers as to where in the codebase the change should be made?
> 
> It would be awesome if such extra information could be added to other issues with the "good first issue" tag too, please :-) thanks in advance!

There's no way to specify how many polygons sketch-to-mesh result should have (i think)

hlorus:
> Hello! Sorry if this should be obvious, but I don't really understand what this feature is about ; could it be possible to add some additional information, given this is tagged as "good first issue"? I'm thinking "current behavior", "expected behavior", and some pointers as to where in the codebase the change should be made?
> 
> It would be awesome if such extra information could be added to other issues with the "good first issue" tag too, please :-) thanks in advance!

Absolutely agree with that, thanks for pointing out! :)

hlorus:
IMO it's better to have this solved by a generic system rather than hardcoding properties that the user might want to change.

There is already an attempt to solve this, see #200.

hlorus:
Closed by fd3b9e3cc30dbbac5a27f3ed840ff1487b40621c

PR #176, opened 2022-06-10
Mesh resolution for Cirle and Arc

BaoNTGH:
- Resolution is per Circle/Arc entity and updatable in the context menu

This is to address issue #58 

Minimum resolution
image
image

Maximum resolution
image
image

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -300,6 +300,16 @@ class SlvsGenericEntity:
         return False
 
 
+class SlvsGenericSegmentedEntity(SlvsGenericEntity):
+    resolution: IntProperty(name="Resolution", default=2, min=2, max=64)
+
+    def bezier_segment_count(self):
+        return self.resolution
+
+    def bezier_point_count(self):
+        return self.bezier_segment_count()
+
+
 # Drawing a point might not include points coord itself but rather a series of virtual points around it
 # so a Entity might refer another point entity and/or add a set of coords
 #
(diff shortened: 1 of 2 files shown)


hlorus (changes requested):
Bumping the amount of bezier segments is probably the wrong approach here, for bezier objects it's probably better to have a minimum amount of control points. As the conversion to mesh happens with blenders native conversion code the resolution of the created mesh probably depends on [this](https://docs.blender.org/api/current/bpy.types.Spline.html#bpy.types.Spline.resolution_u).

This would however mean that it cannot be defined per entity...

Also i had a chat about this topic with @teddyw last week, in short: for now it's ok to define such specific properties for wanted features but generally it's better to have a system that can handle generic properties/attributes that the user can define. Such a system kinda depends on attributes support for curve objects. I do have a WIP branch for that, i will publish it once it's somewhat functional.

BaoNTGH:
> Bumping the amount of bezier segments is probably the wrong approach here, for bezier objects it's probably better to have a minimum amount of control points. As the conversion to mesh happens with blenders native conversion code the resolution of the created mesh probably depends on [this](https://docs.blender.org/api/current/bpy.types.Spline.html#bpy.types.Spline.resolution_u).
> 
> This would however mean that it cannot be defined per entity...
> 
> Also i had a chat about this topic with @teddywaweru last week, in short: for now it's ok to define such specific properties for wanted features but generally it's better to have a system that can handle generic properties/attributes that the user can define. Such a system kinda depends on attributes support for curve objects. I do have a WIP branch for that, i will publish it once it's somewhat functional.

Sound good, you suggested to hold off this PR right?

hlorus:
> > Bumping the amount of bezier segments is probably the wrong approach here, for bezier objects it's probably better to have a minimum amount of control points. As the conversion to mesh happens with blenders native conversion code the resolution of the created mesh probably depends on [this](https://docs.blender.org/api/current/bpy.types.Spline.html#bpy.types.Spline.resolution_u).
> > This would however mean that it cannot be defined per entity...
> > Also i had a chat about this topic with @teddywaweru last week, in short: for now it's ok to define such specific properties for wanted features but generally it's better to have a system that can handle generic properties/attributes that the user can define. Such a system kinda depends on attributes support for curve objects. I do have a WIP branch for that, i will publish it once it's somewhat functional.
> 
> Sound good, you suggested to hold off this PR right?

yep, i would say so

PR #178, opened 2022-06-12
Angle inv (fixes bug 126) ready to go! (fingers crossed)

exegetor:
appears to work beautifully, doesn't cause wild geometric spasms, and it's much more elegant than my first try.
ready to give it a spin?

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -2173,12 +2173,7 @@ def invert_angle_getter(self):
     return self.get("setting", self.bl_rna.properties["setting"].default)
 
 def invert_angle_setter(self, setting):
-#    old_val = self.value  #for the debug printout only
     self["value"] = math.pi - self.value
-#    print("class_defines::invert_angle_setter() - " 
-#    + "   " + str(old_val)
-#    + "   " + str(setting)
-#    + "   " + str(self.value))
     self["setting"] = setting
 
 
@@ -2192,7 +2187,7 @@ class SlvsAngle(GenericConstraint, PropertyGroup):
     value: FloatProperty(
         name=label, subtype="ANGLE", unit="ROTATION", update=update_system_cb
     )
-    setting: BoolProperty(name="Invert", get=invert_angle_getter, set=invert_angle_setter)
+    setting: BoolProperty(name="Measure supplementary angle", get=invert_angle_getter, set=invert_angle_setter)
     draw_offset: FloatProperty(name="Draw Offset", default=1)
     type = "ANGLE"
     signature = ((SlvsLine2D,), (SlvsLine2D,))
@@ -2260,22 +2255,27 @@ class SlvsAngle(GenericConstraint, PropertyGroup):
         return math.degrees(math.acos(x))
 
     def init_props(self):
-        # Set initial angle value to the current angle
+        '''
+        initializes value (angle, in radians),
+            setting ("measure supplimentary angle")
+            and distance to dimension text (draw_offset)
+        '''
+
         line1, line2 = self.entity1, self.entity2
         vec1, vec2 = line1.direction_vec(), line2.direction_vec()
-        angle_std = self._get_angle(vec1, vec2)
-        angle_inv = 180 - angle_std
-        setting = angle_inv < 90
-        self.setting = setting
-        angle = angle_inv if setting else angle_std
+        angle = self._get_angle(vec1, vec2)
+        setting = angle > 90
+        if not setting:
+            angle = 180 - angle
 
-        # Get the radius
         origin = functions.get_line_intersection(
             *functions.line_abc_form(line1.p1.co, line1.p2.co),
             *functions.line_abc_form(line2.p1.co, line2.p2.co),
         )
         dist = max(
-            (line1.midpoint() - origin).length, (line2.midpoint() - origin).length, 0.5
+            (line1.midpoint() - origin).length,
+            (line2.midpoint() - origin).length, 
+            0.5
         )
         self.draw_offset = dist if not setting else -dist
         return math.radians(angle), setting
--- a/class_defines.py
+++ b/class_defines.py
@@ -1,3 +1,4 @@
+from cmath import pi
 import bpy
 import logging
 from bpy.types import PropertyGroup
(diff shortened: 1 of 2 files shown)


exegetor:
Wait, no, it's incomplete...  after creating an Angle, the context menu works great, but if you use the invert checkbox on a new Angle, the setter isn't called, so the geometry jumps.  WIP, then.

exegetor:
<sigh> I guess I need some hints.    The initial pop-up with the angle settings (the one in the lower left) is frustrating me.  I can't figure out how to interact with it in code (like intercepting the signal when "invert" is clicked).   Any pointers?

hlorus:
You should be able to use getter/getters on the operator property as well.
The thing is that it's not really clear what the user expects when creating an angle constraint, e.g. when the user first sets the value he probably wouldn't want to have it changed when adjusting other settings.

exegetor:
Thanks for the tip, everything works now!  Ready for review!
I'd say it's important that the Invert checkbox works consistently whether in the Setup Constraint dialog or in the Context Menu.
As long as it's consistent, users will quickly understand how to use it.

https://user-images.githubusercontent.com/10360161/173665594-9714a34d-6c20-4f13-b6a7-813bc91f1708.mp4

hlorus:
This feels great! The only thing i noticed is that the value in the operator panel can get out of sync with the value which is set on the constraint for some reason.

image

class_defines.py
```diff
@@ -2247,43 +2259,26 @@ def _get_angle(A, B):
...
         line1, line2 = self.entity1, self.entity2
-
         vec1, vec2 = line1.direction_vec(), line2.direction_vec()
         angle_std = self._get_angle(vec1, vec2)
-        angle_inv = self._get_angle_inv(vec1, vec2)
-
-        setting = angle_inv < angle_std
-        angle = angle_inv if setting else angle_std
-
-        self.value = angle
```
exegetor, line 2273: the Operator will soon use the return value to update this, so we don't need it here

operators.py
```diff
@@ -2996,6 +2996,8 @@ def main(self, context):
 
     def fini(self, context, succeede):
         if hasattr(self, "target"):
+            if hasattr(self.target, "inhibit"):
```
exegetor, line 2999: this is odd... the target's Inhibit Flag seems to remain true as soon as we hit a breakpoint in class_defines.py invert_angle_setter() where the update continues to be inhibited (correctly).
But later, using the Context Menu, the Inhibit Flag is correctly False.
I'm not sure why this "works"

exegetor:
So, I got the Operator to stay in sync with the Constraint, but encounterd the issue of BOTH setters being called while the Operator is in control.  Hence the Inhibit Flag on the Constraint.
While the Operator is in control, the Constraint won't adjust the Angle.  Once the Operator shuts down, the Constraint takes over adjusting for the Context Menu Invert.
At initialization Setting and Old_setting are equal.  As soon as you click on Invert, they forever remain opposite.
Bottom line is I'm not confident that this will hold up under serious testing.  I think I'll take few days to recover before hitting this again.   Any thoughts or ideas?

exegetor:
Starting fresh, and simplifying.   The Setter function in class_defines looks completely superfluous.  Checking what happens if I just remove it (and associated artifacts) and rely on the Operator alone.  Expecting to post results of this experiment tomorrow.

hlorus:
I agree it's probably better to keep it simple. IMO having it jump around when twerking from the operator menu isn't as bad as when it jumps once you tweak an already placed constraint.

exegetor:
I'm going to say this is pretty good now, and maybe even good enough.  There is one little detail, though, that causes occasional initial jumping.  The add_angle operator must have old_setting initialized to False, so that a new angle can be correctly inverted (or not) when first displayed.   But if old_setting remains True from a prior state, add_angle can improperly invert a new angle.   I haven't been able to find a place to initialize old_setting so that it's set to False as soon as you hit the "angle" button.
As you said, it's not the end of the world if the operator causes some jumping.  Once add_angle closes, SlvsAngle takes over and handles things well.
I left the debugging print statements because I think you'll want some visibility to check out what's happening.
Whew.  What a nightmare of stateful classes messing with other stateful classes. ;)   Fingers crossed you don't find anything terribly wrong this time!

class_defines.py
```diff
@@ -2180,7 +2192,7 @@ class SlvsAngle(GenericConstraint, PropertyGroup):
     value: FloatProperty(
         name=label, subtype="ANGLE", unit="ROTATION", update=update_system_cb
     )
-    setting: BoolProperty(name="Invert", update=update_system_cb)
+    setting: BoolProperty(name="Invert", get=invert_angle_getter, set=invert_angle_setter)
```
BaoNTGH, line 2195: "setting" name is quite confusing. Maybe explicitly name it inverted or should_invert or similar?

operators.py
```diff
@@ -3055,13 +3077,24 @@ class VIEW3D_OT_slvs_add_angle(
     value: FloatProperty(
         name="Angle", subtype="ANGLE", unit="ROTATION", options={"SKIP_SAVE"}
     )
-    setting: BoolProperty(name="Invert")
+    old_setting: BoolProperty(default = False)
```
BaoNTGH, line 3080: Maybe should SKIP_SAVE this one?

class_defines.py
exegetor: True, but I left it alone since the Distance constraint has "setting" that means "Flip."  There is shared infrastructure.  The parent class uses "setting" without knowing what it means.   Having a more descriptive name would require some re-engineering.

operators.py
exegetor: good call

hlorus:
Hmm i still have the feeling that the old_setting wouldn't really be used, in initialize_constraint() the operator's value and setting properties are set based on the initialized values from the constraint, this currently triggers the setters on the operator however it should be possible to avoid that by directly setting the id properties e.g. self["setting"] = True

operators.py
```diff
@@ -3055,13 +3077,24 @@ class VIEW3D_OT_slvs_add_angle(
     value: FloatProperty(
         name="Angle", subtype="ANGLE", unit="ROTATION", options={"SKIP_SAVE"}
     )
-    setting: BoolProperty(name="Invert")
+    old_setting: BoolProperty(default = False, options={"SKIP_SAVE"})
+    setting: BoolProperty(name="Invert", default = False, get=invert_angle_getter, set=invert_angle_setter)
```
BaoNTGH, line 3081: Instead of using getter/setter I tested with using update and it works great.

If you want to continue with getter/setter, maybe you could use self.properties.is_property_set("settings") to detect the first time?

hlorus:
Awesome, this works like a charm! Thanks a lot for the effort @exegetor!

Issue #38, opened 2022-04-25
Switch active tool when entering & existing sketching mode


teddywaweru:
Could it be possible to switch the active tool that is currently selected to the Solvespace Select Tool when adding or editing the sketch. Then shifting back again once the sketch mode is completed. Obviously most of the typical tools wouldn't be active in the sketch mode & vice versa.

Additionally, possibility of collapsing/hiding the Sketch tool set once the sketch mode has been existed.

hlorus:
That would be a nice addition indeed

PR #179, opened 2022-06-13
Switch active tool when entering & existing sketching mode

BaoNTGH:
Addressing issue #38


hlorus:
This seems to work great, however i'm not sure if it's always the expected behavior, maybe the tool shouldn't be changed when there's no convert type set.

BaoNTGH:
> This seems to work great, however i'm not sure if it's always the expected behavior, maybe the tool shouldn't be changed when there's no convert type set.

I thought it should regardless there's no covert type right? When user is not in sketch mode, they're most likely need to do other Blender action. Without switching, it's like Blender is not working at all until manually switch.

PR #183, opened 2022-06-13
Support constraint visibility toggling

BaoNTGH:
- With complex sketches, constraints are sometimes interfering into
  designing/reading process.
- Show/hide individual constraint in the constraint browser
- Also provide 3 button shortcuts to show/hide/toggle all contraints at
  once

image

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -1676,6 +1676,7 @@ curve = (SlvsCircle, SlvsArc)
 
 class GenericConstraint:
     failed: BoolProperty(name="Failed")
+    visible: BoolProperty(name="Visible", default=True, update=functions.update_cb)
     signature = ()
 
     def needs_wp(args):
@@ -1733,12 +1734,17 @@ class GenericConstraint:
                 continue
             _update(prop_name)
 
+    def is_visible(self, context):
+        if hasattr(self, "sketch"):
+            return self.sketch.is_visible(context) and self.visible
+        return self.visible
+
     def is_active(self, active_sketch):
         if not hasattr(self, "sketch"):
             return not active_sketch
 
         show_inactive = not functions.get_prefs().hide_inactive_constraints
-        if show_inactive and self.sketch.visible:
+        if show_inactive and self.is_visible():
             return True
 
         return self.sketch == active_sketch
--- a/gizmos.py
+++ b/gizmos.py
@@ -124,6 +124,8 @@ class VIEW3D_GT_slvs_constraint(ConstraintGizmo, Gizmo):
 
     def draw(self, context):
         constraint = self._get_constraint(context)
+        if not constraint.visible:
+            return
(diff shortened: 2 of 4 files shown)


BaoNTGH:
@TheMakerTales @hlorus maybe we should introduce hotkeys for these but I'm very bad at this. If you guys have any suggestion, I'll add in.

class_defines.py
```diff
@@ -1733,12 +1734,20 @@ def _update(name):
                 continue
             _update(prop_name)
 
+    def is_visible(self, context):
+        if self.origin:
+            return context.scene.sketcher.show_origin
```
hlorus, line 1739: This looks like a copy/paste leftover

gizmos.py
```diff
@@ -225,6 +230,8 @@ def draw(self, context):
 
     def draw_select(self, context, select_id):
```
hlorus, line 231: I assume the gizmos which define the test_select function should also check the visibility first.

operators.py
```diff
@@ -3212,6 +3216,66 @@ class VIEW3D_OT_slvs_add_ratio(
...
+def set_all_constraints_visibility(context, visible=True, should_toggle=False):
+    constraint_lists = context.scene.sketcher.constraints.get_lists()
+    for constraint_list in constraint_lists:
+        for constraint in constraint_list:
+            if not hasattr(constraint, "visible"):
+                continue
+            constraint.visible = not constraint.visible if should_toggle else visible
+
+
+class View3D_OT_slvs_toggle_all_constraints_visibility(Operator, HighlightElement):
```
hlorus, line 3228: Would be great to combine the operators and just have an enumproperty for the mode...

hlorus (changes requested):
Awesome work, thanks already!

ui.py
```diff
@@ -235,6 +235,19 @@ def draw(self, context):
...
+            operators.View3D_OT_slvs_hide_all_constraints.bl_idname,
+            text="Hide all",
+        )
+        layout.operator(
+            operators.View3D_OT_slvs_show_all_constraints.bl_idname,
+            text="Show all",
+        )
+        layout.operator(
+            operators.View3D_OT_slvs_toggle_all_constraints_visibility.bl_idname,
+            text="Toggle visibility",
```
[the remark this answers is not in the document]
BaoNTGH: Now thinking again, I agree.

gizmos.py
hlorus: Yeah i wasn't able to actually select it when it's not drawn, strange...

class_defines.py
BaoNTGH: Got it, removed.

operators.py
BaoNTGH: Updated to use operator_enum.

PR #184, opened 2022-06-14
Add sketch in Add Menu

BaoNTGH:
image


hlorus (changes requested):
IMO this is a great addition. I would move the code to ui.py however.

BaoNTGH:
> 

Thanks, moved to ui.py

hlorus:
Great, thanks!

PR #185, opened 2022-06-15
add margin for Distance dimension text (solves Bug 145) - comments addressed

exegetor:
WIP.   Distance constraint looks OK.  Angle and Diameter not implemented yet.

https://user-images.githubusercontent.com/10360161/173745629-c30901cc-6c7c-482a-b42a-945c8136047c.mp4

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -2100,12 +2100,14 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
             layout.prop(self, "draw_offset")
 
 
-    def value_placement(self, context):
+    def value_placement(self, context, margin):
         """location to display the constraint value"""
         region = context.region
         rv3d = context.space_data.region_3d
+
         ui_scale = context.preferences.system.ui_scale
-        coords = self.matrix_basis() @ Vector((0, self.draw_offset * ui_scale, 0))
+        offset = ui_scale * (self.draw_offset + margin)
+        coords = self.matrix_basis() @ Vector((0, offset, 0))
         return location_3d_to_region_2d(region, rv3d, coords)
 
 slvs_entity_pointer(SlvsDistance, "entity1")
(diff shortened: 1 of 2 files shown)


BaoNTGH:
Looks cool, I know it's not ready for review yet but wonder if the text should be rotated to parallel with dimension as well. Seeing other CADs doing that and I feel it's cleaner (not certainly we must do it but just something to call out)

hlorus:
> Looks cool, I know it's not ready for review yet but wonder if the text should be rotated to parallel with dimension as well. Seeing other CADs doing that and I feel it's cleaner (not certainly we must do it but just something to call out)

There's actually a nice discussion about this already, see: https://github.com/hlorus/CAD_Sketcher/discussions/159

exegetor:
Ok, all done.  The Diameter text now "avoids" landing on the center-point. which is a great side-effect!

hlorus:
The margin for the diameter gizmo looks quite big when pulling it outside the circle, see:
image

Also i'm not sure if the margin has to be scaled, would probably work with a static value as well but no strong opinion on that.

hlorus:
It looks like the branch changes some other things as well:
- Always place the text on the outside
- Size helpline overshoot by arrow scale

IMO these changes are good, just wanted to note to keep the overview.

exegetor:
Removing the change to the helplines just so I can bundle it with another helpline upgrade.  (the outside Diameter, opposite side, needs a length limit.
I also increased the margin for Angle.  It looked too tight.
The values can still overlap the lines, but it's kinda a balance.  A margin big enough to always prevent overlap can leave the value hanging off in the distance.  Mostly with Distance constraint.  Rotating the text would fix that, but I'm starting to sweat over CPU cycles, ya know?

hlorus:
Alright, this looks like it's good to go!

hlorus:
Unfortunatley i had to revert the 3d-space based margin part of this PR for now. This was leading to issues when working in small scales.

See the commit for further details:
9bafbba83c41e7e350f30301c5e988e03550a367

Issue #95, opened 2022-05-02
Add shortcut hint to constraint operators


hlorus:

hlorus:
Closed by #190

PR #190, opened 2022-06-16
Display shortcut hints for operators

BaoNTGH:
This is proposal for #95 

- Extract out all the operators' names into a separate module to be
  reused without introducing cycle dependencies. Also put all the names
  in one place
- Original thought was to have each operator defines its own keymap (makes
  more sense in term single responsibility) but due to the nature of key
  mapping where management/cross reference are more important, move them
  to a common place keymaps.py to be shared across.

Multiple shortcuts:
image

Single shortcut:
image

### DIFF
--- a/declarations.py
+++ b/declarations.py
@@ -1,0 +1,82 @@
+# Please keep this file in alphabetical order
+
+from enum import  Enum
+
+class Gizmos(str, Enum):
+    Angle = "VIEW3D_GT_slvs_angle"
+    Constraint = "VIEW3D_GT_slvs_constraint"
+    ConstraintValue = "VIEW3D_GT_slvs_constraint_value"
+    Diameter = "VIEW3D_GT_slvs_diameter"
+    Distance = "VIEW3D_GT_slvs_distance"
+    Preselection = "VIEW3D_GT_slvs_preselection"
+
+class GizmoGroups(str, Enum):
+    Angle = "VIEW3D_GGT_slvs_angle"
+    Constraint = "VIEW3D_GGT_slvs_constraint"
+    Diameter = "VIEW3D_GGT_slvs_diameter"
+    Distance = "VIEW3D_GGT_slvs_distance"
+    Preselection = "VIEW3D_GGT_slvs_preselection"
+
+class Menus(str, Enum):
+    Sketches = "VIEW3D_MT_sketches"
+
+class Operators(str, Enum):
+    AddAngle = "view3d.slvs_add_angle"
+    AddArc2D = "view3d.slvs_add_arc2d"
+    AddCircle2D = "view3d.slvs_add_circle2d"
+    AddCoincident = "view3d.slvs_add_coincident"
+    AddDiameter = "view3d.slvs_add_diameter"
+    AddDistance = "view3d.slvs_add_distance"
+    AddEqual = "view3d.slvs_add_equal"
+    AddHorizontal = "view3d.slvs_add_horizontal"
+    AddLine2D = "view3d.slvs_add_line2d"
+    AddLine3D = "view3d.slvs_add_line3d"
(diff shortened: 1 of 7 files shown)


hlorus:
Really great addition and cleanup! I like the the hint to be displayed on top, right after the label.
What's currently a bit odd is that it's different from the workspacetool's tooltip, would be great to have them consistent, it should probably be possible to use the same function for both places.

BaoNTGH:
> Really great addition and cleanup! I like the the hint to be displayed on top, right after the label. What's currently a bit odd is that it's different from the workspacetool's tooltip, would be great to have them consistent, it should probably be possible to use the same function for both places.

Just to clarify that, you would prefer the single short cut screen above right? (It's possible to update the tool's hint to have similar as above single short cut.)

Also, for multiple shortcuts, would you be ok with above or also update to:
**Add Distance constraint (Alt + D, Alt + V, Alt + H)**

hlorus:
> > Really great addition and cleanup! I like the the hint to be displayed on top, right after the label. What's currently a bit odd is that it's different from the workspacetool's tooltip, would be great to have them consistent, it should probably be possible to use the same function for both places.
> 
> Just to clarify that, you would prefer the single short cut screen above right? (It's possible to update the tool's hint to have similar as above single short cut.)

Yeah, the same as in the screenshot you've posted.


> Also, for multiple shortcuts, would you be ok with above or also update to:
> **Add Distance constraint (Alt + D, Alt + V, Alt + H)**

Not sure about that, IMO both is okay. The only problem i see is that there's a maximum count of characters for the tooltip and it will get clipped if it's too long, so maybe better to display them inline.

BaoNTGH:
> > > Really great addition and cleanup! I like the the hint to be displayed on top, right after the label. What's currently a bit odd is that it's different from the workspacetool's tooltip, would be great to have them consistent, it should probably be possible to use the same function for both places.
> > 
> > Just to clarify that, you would prefer the single short cut screen above right? (It's possible to update the tool's hint to have similar as above single short cut.)
> 
> Yeah, the same as in the screenshot you've posted.
> 
> 
> > Also, for multiple shortcuts, would you be ok with above or also update to:
> > **Add Distance constraint (Alt + D, Alt + V, Alt + H)**
> 
> Not sure about that, IMO both is okay. The only problem i see is that there's a maximum count of characters for the tooltip and it will get clipped if it's too long, so maybe better to display them inline.
> 
> 

Ideally, I think one operator should do just one thing. It'll be less confusing. For example, direct distance, vertical distance and horizontal distance as there are actually different constraints IMO. It's clearer for user to use hot key as well. But I'll defer that for future discussion. Will update to be inlined for consistency for now

hlorus:
There is really just one distance constraint now with the option to be aligned or not. So this currently communicates the implementation quite well.
It might be possible to check through properties when searching for a keymap item and then expose the distance constraint with three buttons but not sure if it's really worth the effort...

BaoNTGH:
> There is really just one distance constraint now with the option to be aligned or not. So this currently communicates the implementation quite well. It might be possible to check through properties when searching for a keymap item and then expose the distance constraint with three buttons but not sure if it's really worth the effort...

Addressed all concerns. Took a chance to do much more cleaning up to reduce dependencies between modules.

image
image

hlorus:
Great, the only thing that's kinda missing is support for global keymaps, see: 0b22b0ec69fe8bf82452efbe14d7d65c14d677b9.

hlorus:
As this PR is kinda getting difficult to review i'm gonna merge this already.

BaoNTGH:
> Great, the only thing that's kinda missing is support for global keymaps, see: 0b22b0ec69fe8bf82452efbe14d7d65c14d677b9.

Yup, that could come later. There are a little more clean up to centralize all the key map at one place

PR #192, opened 2022-06-19
add dimension line width setting to prefs (issue 143)

exegetor:
In case you think this is "good enough".    Separate widths for the leader lines vs witness lines will require much more thinking

### DIFF
--- a/gizmos.py
+++ b/gizmos.py
@@ -227,6 +227,7 @@ class ConstraintGizmoGeneric(ConstraintGizmo):
         self._update_matrix_basis(constr)
 
         self._create_shape(context, constr)
+        gpu.state.line_width_set(functions.get_prefs().gizmo_width)
         self.draw_custom_shape(self.custom_shape)
 
     def draw_select(self, context, select_id):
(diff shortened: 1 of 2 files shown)


hlorus:
It would be great if someone could check if this works on mac, then we can add this.

TheMakerTales:
@hlorus @exegetor 
Just tested this on a mac and it does not work =/

exegetor:
@TheMakerTales that's what we expected, unfortunately.
As this PR comes out of a user request, I'll spend a little time looking into how this could be made to work on Mac.
I'm not optimistic about that...
As an alternative, what about making it a Win/Linux feature only?  (Mac wouldn't even see the option in Prefs)

hlorus:
> @TheMakerTales that's what we expected, unfortunately.
> As this PR comes out of a user request, I'll spend a little time looking into how this could be made to work on Mac.
> I'm not optimistic about that...
> As an alternative, what about making it a Win/Linux feature only?  (Mac wouldn't even see the option in Prefs)
> 

I wouldn't invest too much time into a solution that doesn't support all platforms...
Better try to draw lines with filled rectangles, then we can use an arbitrary thickness. Afaik the LINES_TRIS type of new_custom_shape is used to draw surfaces, see: https://docs.blender.org/api/current/bpy.types.Gizmo.html#bpy.types.Gizmo.new_custom_shape

PR #194, opened 2022-06-20
Panel heirachy change Issue#169

init-pixel:
Added type hints to assist with code completion.
Added a second blank line between classes where necessary for consistency.
Added a mix-in class, 'VIEW3D_PT_sketcher_base', to remove unnecessary specification of ui attributes (space, region, category), and Panel inheritance. Mix-in also used to register all panels based on them being subclasses rather than it being necessary to manually specify them.
Split add constraint operators and debug params out into their own panels,  assigned new id names in declarations.

As a side note, can we rename 'declarations.py' to 'bl_idnames.py' or 'id_names.py', as I don't think the current name communicates the file's purpose well.

### DIFF
--- a/declarations.py
+++ b/declarations.py
@@ -1,6 +1,6 @@
 # Please keep this file in alphabetical order
+from enum import Enum
 
-from enum import  Enum
 
 class Gizmos(str, Enum):
     Angle = "VIEW3D_GT_slvs_angle"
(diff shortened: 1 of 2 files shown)


hlorus:
This looks good :+1: 

> As a side note, can we rename 'declarations.py' to 'bl_idnames.py' or 'id_names.py', as I don't think the current name communicates the file's purpose well.
I don't have a problem with the current name but also wouldn't care if it's renamed, maybe @BaoNTGH has an opinion on that as he named it in the first place...

BaoNTGH:
> This looks good 👍
> 
> > As a side note, can we rename 'declarations.py' to 'bl_idnames.py' or 'id_names.py', as I don't think the current name communicates the file's purpose well.
> > I don't have a problem with the current name but also wouldn't care if it's renamed, maybe @BaoNTGH has an opinion on that as he named it in the first place...

The idea is it's not just used for idnames, others could be used to be declared (mostly declarations, not much implementation) so it could be shared across.

I'm thinking to clean up more like all the enum for solver states, alignments... More like for house keeping, no big deal in feature.

PR #195, opened 2022-06-20
limit length of opposite-side diameter leader / scale overshoot:  comments addressed

exegetor:
These improvements are needed (IMO) to make my earlier changes more correct.

First, making the Arrow Size user-settable suddenly made the overshoot constant seem out of proportion.  I've roughly approximated one arrowhead length for the overshoot.

Second, the Diameter constraint (outside mode) should not let the opposite-side leader get out of hand.  I've set a limit of 4 arrowheads  (BTW a great measurement unit!)

### DIFF
--- a/gizmos.py
+++ b/gizmos.py
@@ -263,6 +263,7 @@ def draw_arrow_shape(target, shoulder, width, is_3d=False):
 def get_overshoot(scale, dir):
     if dir == 0:
         return 0
+    #use factor of 0.005 for one-half arrowhead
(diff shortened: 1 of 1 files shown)


hlorus:
IMO a much needed improvement! I would suggest some minor tweaks though:
- Use a static value for the opposite-side diameter overshoot (e.g. 2*arrow_length). There's no need to have it sized based on the offset.
- Reduce general overshoot a bit, maybe just half the arrow size?

Issue #167, opened 2022-06-08
Adjust pip usage to prevent interfering with numpy.


TheMakerTales:

hlorus:
Afaik this has already been addressed, closing.

PR #198, opened 2022-06-21
improved pip handling (issue #167), type hints, and cleanup

init-pixel:
Modifications to `functions.py`
1. Added `--no-deps` flag to `install_package` function. Doesn't affect current deps installation but if additional packages are required in future they could potential interfere with Blender's numpy version.
2. Added `refresh_path` function, and call inside `install_package` which should reduce necessity of Blender restarts after package installation. Admittedly I've never been able to experience the issue but in theory this should be a fix.
3. General cleanup through type hints and consistent line spaces.
4. Imports cleanup. Grouped by source. Removed mid file imports. Replaced `2pi` definition with built-in `tau` from `math`. There's other instances where functions could be replaced by built-ins (or bpy, numpy) equivalents for better performance which I'll look at in future.

### DIFF
--- a/functions.py
+++ b/functions.py
@@ -1,7 +1,25 @@
-import bpy
-from . import global_data
-from mathutils import Vector, Matrix
+from collections import deque
+import math
+from math import pi, cos, sin, tau
 import subprocess
+import re
+import site
+from importlib import reload
+from typing import Any, List, Tuple, Union
+
+import bpy
+import bmesh
+from bpy.types import Context, RegionView3D, Object
+from bpy_extras.view3d_utils import (
+    location_3d_to_region_2d,
+    region_2d_to_location_3d,
+    region_2d_to_vector_3d,
+    region_2d_to_origin_3d,
+)
+from mathutils import Vector, Matrix
+from mathutils.bvhtree import BVHTree
+
+from . import global_data
 
 
 def get_prefs():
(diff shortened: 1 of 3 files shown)


functions.py
```diff
@@ -156,136 +198,129 @@ def coords_arc_2d(
...
+    view_vector = region_2d_to_vector_3d(region, rv3d, coords)
+    ray_origin = region_2d_to_origin_3d(region, rv3d, coords)
 
     # view vector needs to be scaled and translated
     end_point = view_vector * context.space_data.clip_end + ray_origin
     return ray_origin, end_point
 
 
-def nearest_point_line_line(p1, d1, p2, d2):
+def nearest_point_line_line(p1: float, d1: float, p2: float, d2: float) -> Vector:
```
hlorus, line 251: These arguments are vectors, not floats.

functions.py
```diff
@@ -135,9 +170,16 @@ def coords_circle_2d(x, y, radius, segments):
...
+    x: float,
+    y: float,
+    radius: float,
+    segments: int,
+    angle=tau,
+    offset: float = 0.0,
+    type="LINE_STRIP",
 ):
-    coords = []
+    # coords = []
```
hlorus, line 181: Can be removed

hlorus (commented):
This needs a rebase, there seems to be a conflict, also a few minor comments in the code otherwise this looks good.

hlorus:
Merged with applied comments. 95682e4b8df24776694622ab4b24b62f706b6d10

PR #201, opened 2022-06-27
dimension Distance and Angle with text outside witness lines --- rebased with all comments addressed

exegetor:
I'll have a video up within 12 hours to show what it looks like and how to work with it.

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -25,7 +25,7 @@ from typing import Union, Tuple, Type, Callable, Mapping
 
 from .shaders import Shaders
 from .solver import solve_system, Solver
-from .functions import unique_attribute_setter
+from .functions import pol2cart, unique_attribute_setter
 from .declarations import Operators
 
 logger = logging.getLogger(__name__)
@@ -2262,8 +2262,9 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
         set=set_distance_value,
     )
     flip: BoolProperty(name="Flip", update=GenericConstraint.update_system_cb)
-    draw_offset: FloatProperty(name="Draw Offset", default=0.3)
     align: EnumProperty(name="Align", items=align_items, update=GenericConstraint.update_system_cb,)
+    draw_offset: FloatProperty(name="Draw Offset", default=0.3)
+    draw_outset: FloatProperty(name="Draw Outset", default=0.0)
     type = "DISTANCE"
     signature = ((*point, *line), (*point, *line, SlvsWorkplane))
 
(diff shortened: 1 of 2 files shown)


class_defines.py
```diff
@@ -2252,6 +2252,8 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
     )
     flip: BoolProperty(name="Flip", update=update_system_cb)
     draw_offset: FloatProperty(name="Draw Offset", default=0.3)
+    draw_outset: FloatProperty(name="Draw Outset", default=0.0)
+    text_inside: BoolProperty(name="Text Inside", default=True)
```
hlorus, line 2256: Not sure why this property would be needed, the text is simply drawn based on the draw_outset, right? You could just define a method to check if the text is inside rather than trying to keep this variable in sync.

exegetor:
You are correct, as always!  
Here's the vid:

https://user-images.githubusercontent.com/10360161/177061909-5f6c3859-3546-4083-94d2-c45edda6fd96.mp4

class_defines.py
```diff
@@ -2402,10 +2400,12 @@ def init_props(self, **kwargs):
         self.value = value
         return value, None
 
+    def text_inside(self, ui_scale):
+        return (ui_scale * abs(self.draw_outset)) < self.value/2
+
```
exegetor, line 2405: I cannot say _why_ this requires ui_scale to track properly.  Tested under windows 11 with display scaling property @ 175%,

exegetor: update: the inconsistency comes from chasing a pre-existing inconsistency.   I'll see if I can expunge the original inconsistency without breaking anything else.  I see what I have to do but it'll take some time to test.  I propose allowing this ugly odd argument to be ---for now.

gizmos.py
```diff
@@ -399,49 +426,73 @@ def _get_helplines(self, context, constr, scale_1, scale_2):
...
+                *draw_arrow_shape(p2, p2 + p2_s, widths[1]),
+                *(self._get_helplines(context, constr, *scales) if not select else ()),
+            )
+        else:
+            outset = constr.draw_outset
+            leader_start = math.copysign(half_angle, outset)
+            leader_end = outset - leader_start
+            coords = (
+                *draw_arrow_shape(p1, p1 - p1_s, widths[0]),
+                *functions.coords_arc_2d(
```
hlorus, line 487: maybe the two two calls to coords_arc_2d could be merged

hlorus (commented):
This looks good, it has to be rebased before merged tho.

gizmos.py
```diff
@@ -1,3 +1,4 @@
+from re import L
```
exegetor, line 1: oops, this is debris

PR #203, opened 2022-06-27
Minor refactors

amrsoll:
Some refactoring to make the code more "python'esque" and hopefully more readable. 

- Function docstrings
- Move ``def function(self, ... )`` inside the appropriate classes
- Break down the data reading from the icons
- Remove some unused functions / unnecessary lines
- move most imports to the start of the file
- A little bit of function typing
- Remove global function `tag_update` -> Classfunction. Less confusing because of blender function with same name

Note: I assumed that `tag_update` does not refer to the `bpy` `tag_update` function. Here I make an explicit reference to `SlvsGenericEntity.tag_update()`

I have done minor testing to see if I didn't break anything major

N.B. I would recommend using a develop branch for people to create PRs against :) just for sanity check before commiting multiple PRs into master

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -10,6 +10,7 @@ from bpy.props import (
     FloatVectorProperty,
     EnumProperty,
     StringProperty,
+    IntVectorProperty,
 )
 
 from . import functions, preferences
@@ -19,6 +20,8 @@ from . import global_data
 
 from bpy_extras.view3d_utils import location_3d_to_region_2d
 import math, mathutils
+from mathutils import Vector, Matrix
+from typing import Union, Tuple, Type, Callable, Mapping
 
 from .shaders import Shaders
 from .solver import solve_system, Solver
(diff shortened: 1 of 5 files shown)


operators.py
```diff
@@ -743,7 +743,6 @@ def register_properties(cls):
         for a in annotations.keys():
             if hasattr(cls, a):
                 raise NameError("Cannot register implicit pointer properties, class {} already has attribute of name {}".format(cls, a))
-        setattr(cls, "__annotations__", annotations)
```
amrsoll, line 746: `annotations` was never modified

testing/utils.py
```diff
@@ -45,8 +45,6 @@ class Sketch2dTestCase(BgsTestCase):
     def setUpClass(cls):
         super().setUpClass()
 
-        from mathutils import Quaternion
```
amrsoll, line 48: unused

class_defines.py
```diff
@@ -660,12 +662,9 @@ def normal(self):
 ]
 
 
-def hide_sketch(self, context):
-    if self.convert_type != "NONE":
-        self.visible = False
```
amrsoll, line 665: unused

hlorus:
LGTM, thanks for the PR!

PR #206, opened 2022-06-28
VSCodium/VSCode: Remote debugger

amrsoll:
Adds config files for Codium, / VS Code to
- start Blender with the CADSketcher add on
- Create a remote debug session in Blender
- Connect to said remote debugger

Introduces optional dependency to `colorama` for coloring stdout if it's on a terminal

I will make a short write-up of the steps to set the dev env up for linux (requires to add a symlink of the CAD_Sketcher repo to the Blender add-on folder)

### DIFF
--- a/.vscode/debugpy_port.py
+++ b/.vscode/debugpy_port.py
@@ -1,0 +1,19 @@
+import debugpy
+
+try:
+    from colorama import Fore
+    YELLOW, GREEN, RED, RESET = Fore.YELLOW, Fore.GREEN, Fore.RED, Fore.RESET
+except ImportError:
+    YELLOW, GREEN, RED, RESET = [""]*4
+
+# https://code.visualstudio.com/docs/python/debugging#_debugging-by-attaching-over-a-network-connection
+
+# 5678 is the default attach port in the VS Code debug configurations. Unless a host and port are specified, host defaults to 127.0.0.1
+debugpy.listen(5678)
+print(f"{YELLOW}Waiting for debugger attach{RESET}")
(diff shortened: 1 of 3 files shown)


amrsoll:
Screenshot of use-case

image

hlorus:
I'm not sure if such config files should be stored in the repository or better just be ignored. I would propose to merge this and if contributors are having issues with this we can add the files to .gitignore then.

hlorus:
Seems to work great on my end so far, awesome!

amrsoll:
I forgot to post before you merged the PR :sweat_smile: 

> I'm not sure if such config files should be stored in the repository or better just be ignored.

Usually they are stored in the repo, unless the maintainer is against having IDE-specific integration of their project.

> if contributors are having issues with this we can add the files to .gitignore then.

VS Code/Codium shows a warning when it detects this folder in a new project and asks the developer if they want to activate it.

In this case, it creates 2 new debugger possibilities; both allow the use of line breaks and remote debugger inside a Blender process.
-  Pytest+debugger : Runs the tests with Blender running background (no gui)
- Interactive+debugger: Starts Blender with the add-on

As pointed out on Discord, it would seem most people are perfectly happy with the "Blender Development" plugin, I simply have had shortcomings with it.

hlorus:
> VS Code/Codium shows a warning when it detects this folder in a new project and asks the developer if they want to activate it.
Then this shouldn't be a problem anyway if other contributers can just opt out, that's great.
What's not nice currently is that these files are also downloaded by users but that's probably another topic.

PR #207, opened 2022-06-29
WIP (one bug known) add circle-line and circle-point distance constraint

exegetor:
i could not find a native py-slvs function for this constraint, so i faked it with an invisible construction line from the circle's centerpoint to the target and a point on the circle and a point on the target line (if applicable).

Constraints are applied to tie it all together.

The generic constraint now has lists for temporary, subordinate entities and constraints.  They will be deleted when the primary constraint is deleted.

HOWEVER, THERE IS STILL ABUG that leaves one subordinate constraint alive after its parent constraint is deleted.
Anybody who spots the bug wins a gold star!

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -2254,12 +2254,11 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
     draw_offset: FloatProperty(name="Draw Offset", default=0.3)
     align: EnumProperty(name="Align", items=align_items, update=update_system_cb,)
     type = "DISTANCE"
-    signature = ((*point, *line), (*point, *line, SlvsWorkplane))
+    signature = ((*point, *line, SlvsCircle), (*point, *line, SlvsWorkplane))
 
     @classmethod
     def get_types(cls, index, entities):
         e = entities[1] if index == 0 else entities[0]
-
         if e:
             if index == 1 and e.is_line():
                 # Allow constraining a single line
@@ -2346,7 +2345,9 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
             return Matrix()
 
         e1, e2 = self.entity1, self.entity2
-        if e1.is_line():
+        if e1.is_curve():
+            return Matrix()
+        elif e1.is_line():
             e1, e2 = e1.p1, e1.p2
 
         sketch = self.sketch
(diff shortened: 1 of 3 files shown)


class_defines.py
```diff
@@ -2346,7 +2347,9 @@ def matrix_basis(self):
             return Matrix()
 
         e1, e2 = self.entity1, self.entity2
-        if e1.is_line():
+        if e1.is_curve():
```
exegetor, line 2350: expert eyes needed to see if this looks funny.  My Matrix-fu is weak.

hlorus:
Hey before proceeding note that creating temporary blender data-blocks is not quite the right approach as it's redundant.

The solvesys is temporary anyway so it's possible to add multiple constraints inside a create_slvs_data function and simply return the handles for every added constraint. This is already used a few times, an example would be the alinged mode of the distance constraint which implicitly adds horizontal/vertical constraints.

exegetor:
Difficulties with the approach you describe.

1)  init_props(), which sets the Distance, is called before create_slvs_data().  It therefore needs to calculate the length of the segment between point A on the circle and point B, the perpendicular projection of A on the target line.  This duplicates calculation.

2) the gizmo needs to have A and B also.  Recalculate?

3) the matrix basis needs to be based on A and B as SlvsPoint2d, when the constraint is between SlvsCircle and SlvsLine.  Again with the recalculations.

4)  even with all this done, now the Circle and Line can't be dragged around.  Something is causing the system to "stick"

I've done all these changes (the matrix basis is not quite correct; the helper lines are spawned at an odd angle) but it looks unwise.  The geometry I create on-the-fly might be a nasty hack but it nicely fits into the infrastructure already in place.  Things "just work" out of the box, so to speak.    What am I missing?  If we rely on create_slvs_data() as you suggest, how do we get access to things we need elsewhere?  and how do we get the system to update on dragging elements around?

hlorus:
> 1. init_props(), which sets the Distance, is called before create_slvs_data().  It therefore needs to calculate the length of the segment between point A on the circle and point B, the perpendicular projection of A on the target line.  This duplicates calculation.
It's the responibility of init_props to guess the correct values when creating a constraint based on the current setup of the entities. This will be called at creation time only. If there's duplicated code a function can be created and shared.

> 2\. the gizmo needs to have A and B also.  Recalculate?
Not really, the gizmo just gets the matrix and distance i think.

> 4\. even with all this done, now the Circle and Line can't be dragged around.  Something is causing the system to "stick"
This is usually because the system is overconstrained or the solversystem gets wrong information like wrong arguments order. Quite difficult to help without seeing the code.

exegetor:
New PR is up now, implementing the whole thing thru solvesys, as requirested

PR #212, opened 2022-07-02
Dist from curve (solvesys version)

exegetor:
I believe this is the droid you were looking for.
There are two things left to do tomorrow:
First, the Horizontal/Vertical alignment is not correct.  The helplines are drawn slightly off the target.
Second, when adjusting the distance via the context menu, negative values are impossible.  It seems as if the circle and line should pass thru each other rather than bounce off.  Is this what the "Flip" feature is supposed to correct for?
Third, Arcs will require special treatment.  I figure the Arc will act like the Circle within its domain, but outside that, the appropriate arc endpoint becomes one end of a Point->Line constraint.

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -2383,8 +2383,15 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
             else:
                 assert(isinstance(e2, SlvsPoint2D))
                 p2 = e2.co
-            vec = (p2 - centerpoint) / (p2 - centerpoint).length
-            p1 = centerpoint + (e1.radius * Vector(vec))
+            if (p2 - centerpoint).length > 0:
+                vec = (p2 - centerpoint) / (p2 - centerpoint).length
+                p1 = centerpoint + (e1.radius * Vector(vec))
+            else:
+            # This is a curve->line where the centerpoint of the curve is
+            # coincident with the line.  By reassigning p1 to an endpoint
+            # of the line, we avoid p1=p2 errors and the result is
+            # (correctly) an invalid constraint
+                p1 = e2.p1.co
         elif e1.is_line():
             # reframe as point->point and continue
             e1, e2 = e1.p1, e1.p2
@@ -2393,11 +2400,10 @@ class SlvsDistance(GenericConstraint, PropertyGroup):
             assert(isinstance(e1,SlvsPoint2D))
             p1 = e1.co
 
-
-        # NOTE: p2 is still not defined if e1 is point
         if type(e2) in point_2d:
-        # this includes "Line Length" (became Point->Point)
-        # and Curve -> Point
+        # this includes "Line Length" (now point->point)
+        # and curve -> point
+            p2 = e2.co
             if not align:
                 v_rotation = p2 - p1
             else:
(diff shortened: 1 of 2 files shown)


class_defines.py
```diff
@@ -2276,6 +2277,8 @@ def needs_wp(self):
 
     def use_flipping(self):
         # Only use flipping for constraint between point and line/workplane
+        if self.entity1.is_curve:
```
hlorus, line 2280: is_curve is a function, has to be called

class_defines.py
```diff
@@ -2345,35 +2358,72 @@ def matrix_basis(self):
...
-            end = line.p2.co - orig
-            angle = functions.range_2pi(math.atan2(end[1], end[0])) + math.pi / 2
-
-            mat_rot = Matrix.Rotation(angle, 2, "Z")
-            p1 = e1.co - orig
-            v_translation = (p1 + p1.project(end)) / 2 + orig
+        elif e2.is_line():
+            if e1.is_curve():
+                if not align:
+                    v_rotation = p2 - p1
```
hlorus, line 2412: This vector could be zero, that case should be handled somehow otherwise we run into errors later

class_defines.py
```diff
@@ -2383,6 +2433,14 @@ def init_props(self, **kwargs):
         e1, e2 = self.entity1, self.entity2
         if e1.is_line():
             value = e1.length
+        elif type(e1) in curve:
+            centerpoint = e1.ct.co
+            if isinstance(e2, SlvsLine2D):
+                endpoint, whatever = intersect_point_line(centerpoint, e2.p1.co, e2.p2.co)
```
hlorus, line 2439: Usually "_" is used as a name for variables that aren't actually used, "whatever" as a name might look like it's actually used somewhere

hlorus (commented):
Besides the points you've mentioned this already works great.

class_defines.py
exegetor: fixed

class_defines.py
```diff
@@ -2345,35 +2358,72 @@ def matrix_basis(self):
...
+        #   arc      line
+        #   circle   point
+        #   circle   line
+
+        # set p1 and p2
+        if e1.is_curve():
+            # reframe as point->point and continue
+            centerpoint = e1.ct.co
+            if e2.is_line():
+                p2, whatever = intersect_point_line(centerpoint, e2.p1.co, e2.p2.co)
```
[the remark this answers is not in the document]
exegetor: fixed

class_defines.py
exegetor: fixed

hlorus:
Something might have gotten wrong when updating, at least i don't see the changes yet that you've marked as resolved...

class_defines.py
```diff
@@ -2276,6 +2277,8 @@ def needs_wp(self):
 
     def use_flipping(self):
         # Only use flipping for constraint between point and line/workplane
+        if self.entity1.is_curve():
+            return False
```
exegetor, line 2281: still need to get Flip working.  Hard to get Circle and Line to switch sides without this.

hlorus:
@exegetor I've pushed a fix to the dist_from_curve branch regarding the problem with the helplines. Would you mind giving it a quick test?

hlorus:
Merged in: https://github.com/hlorus/CAD_Sketcher/commit/07362c2affc993c99e873137f88827c3ba1cb3d6

PR #221, opened 2022-07-07
add Preferences.show_mesh (recolves discussion 113)  COMMENTS ADDRESSED

exegetor:
Not sure how often I'd want to edit a sketch and see its mesh, but... users have asked for it, it doesn't chew up the CPU, 
 its easy to implement, and I had a few spare hours

### DIFF
--- a/base/preferences.py
+++ b/base/preferences.py
@@ -142,6 +142,7 @@ class Preferences(AddonPreferences):
         name='Angle Precision', min=0, max=5, default=0,
         description="Angle decimal precision")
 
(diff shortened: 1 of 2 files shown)


hlorus (commented):
Lgtm, the name is a bit misleading as it also applies to curve objects. Maybe something like:
"use_hide_objects"
"auto_hide_objects"

exegetor:
should be good now.   I'm partially back from a family emergency.  I say partially, because it will linger a bit.

hlorus:
Great to see you're back, hope everything is ok. I will have a look at the updated PR tomorrow.

hlorus:
Merged with minor adjustments

PR #222, opened 2022-07-11
Dependecy reporting and cleanup

init-pixel:
Added reporting of dependencies preventing entity deletion to `View3D_OT_slvs_delete_entity` and a small function to display a popup to perform the reporting.
Replaced the mean function defined in `class defines` with the appropriate built-in.
Fixed a number of instances where the presence of None was being checked through equality rather than identity.
General formatting for consistence with line spacing and other such small adjustments to be more in line with PEP8.


hlorus:
Merged with minor adjustments

PR #223, opened 2022-07-14
Cleanup init

hlorus:
An attempt to clean up the registeration flow of the addon and add a future proof structure

- Add register packages which register all blender classes
    - Base: Registers basic addon (theme, preferences, install operator)
    - Core: Registers everything else (operators, panels, menus, property groups), this will still need to be ported over
- Add utilities package which holds stuff that doesn't need to directly be registered
- Remove importlib.reload functionality in favor of sys.modules purging to allow hot-reload
- Pack ressources into subfolder (icons, logos, presets)

The planned addon structure looks like this:
├── base
│   ├── __init__.py
│   ├── install.py
│   ├── preferences.py
│   └── theme.py
├── core 
│   ├── __init__.py
│   ├── properties
│   │   ├── __init__.py
│   │   ├── point_3d.py
│   │   └── line_2d.py
│   ├── operators
│   │   ├── __init__.py
│   │   ├── point_3d.py
│   │   └── line_2d.py
│   └── ui
│   │   ├── __init__.py
│   │   ├── panel_main.py
│   │   ├── panel_debug.py
│   │   ├── gizmo_generic.py
│   │   ├── tool_select.py
│   └── keymap.py

├── utilities
│   ├── __init__.py
│   ├── geometry.py
│   ├── units.py
│   └── data_handling.py

├── __init__.py
├── constants.py
├── register_full.py
├── shaders.py
├── solver.py
├── converters.py
└── versioning.py


mkingsnorth:
Based on my experience of the code so far I had the following thoughts:

- **The difference between base and core**:  I understand that base is for 'basic' add-on functions and the 'core' is for everything else, there might be a risk that code could get bloated again in the future inside 'core'.  What else did you consider?  For instance, KIT OPS has an 'interface' package at the top level that just has the main operators definitions in it and the panel.  I'm not saying that is exactly right, but it was good as a new developer to see that code first.  I like the 'operators' and 'ui' packages within 'core'...perhaps you could consider arranging those packages so that they are at a first level. 'preferences', 'theme' and 'install' py files perform different functions so perhaps they should be at the top level or split between packages.
- **Properties package**: I wonder whether the 'properties' package is more about the Data Model, and perhaps could be renamed to 'model'.  Have you considered using any Design Patterns in the refactor, such as Model View Controller (in this case, Model = 'properties', View = 'ui', Controller = 'operator', so it is close)?
- **Utilities Package is ok:** utilities package always seemed to be needed :)
- **Top Level py files:** I feel like the things at the top level (constants, shaders, solver in particular) are still part of the core and need to be treated as such.  Is the solver.py package more of an wrapper to the solver library?
- **Stateful Operator and gpl code:** Is the StatefulOperator code part of the utility package in this refactor?  Are all the gpl rendering functions embedded in that or is it all part of gizmos?

hlorus:
> Based on my experience of the code so far I had the following thoughts:
> 
> - **The difference between base and core**:  

The main thought behind the two submodules is to communicate the registration flow of the addon. As it heavily depends on the solver module the registration is split into two parts, this prevents access when the solver isn't accessible.
I absolutely agree that the core module gets bloated as almost any operator will live there. Having the core subpackages live in the root directory is IMO great, the only downside here is that you cannot easily tell which packages are part of the registration.

> - **Properties package**: 

Sounds great. IMO the name itself could be a bit more descriptive tho. Maybe something like data_model, structure, architecture?


> - **Top Level py files:** 
They aren't not part of the registration flow tho... Again not exactly sure how this could be communicated.


> - **Stateful Operator and gpl code:**
 Is the StatefulOperator code part of the utility package in this refactor? 

Either that or just as a package in the root directory.

> Are all the gpl rendering functions embedded in that or is it all part of gizmos?

Haven't really looked at this. I assume entity drawing would still be defined inside the propertygroups definitions, so in core->properties->some_entity.py. Same for the gizmos but this could probably be improved...

mkingsnorth:
> The main thought behind the two submodules is to communicate the registration flow of the addon

Question: Why is the registration flow need to be the focus of the refactoring effort?  Do the classes etc need to be in a certain order to be automatically registered?  I ask because it seems restrictive that the code needs to be ordered in a specific way in order to be registered correctly.

hlorus:
So with your comments addressed the structure should be something like this, with the registration.py defining which module's register function are being called:
├── base
│   ├── __init__.py
│   ├── install.py
│   ├── preferences.py
│   └── theme.py

├── data_model
│   ├── __init__.py
│   ├── point_3d.py
│   └── line_2d.py

├── operators
│   ├── __init__.py
│   ├── point_3d.py
│   └── line_2d.py

├── interface
│   ├── __init__.py
│   ├── panel_main.py
│   ├── panel_debug.py
│   ├── gizmo_generic.py
│   ├── tool_select.py

├── utilities
│   ├── __init__.py
│   ├── geometry.py
│   ├── units.py
│   ├── solver.py
│   ├── converters.py
│   └── data_handling.py

├── stateful_operator
│   ├── __init__.py
│   ├── logic.py
│   ├── integration.py
│   └── tool.py

├── __init__.py
├── constants.py
├── registration.py
├── shaders.py
├── keymap.py
└── versioning.py

mkingsnorth:
This looks like a great improvement in my opinion.  Without delving into the codebase further, this looks like a good start and then further improvements can be made over time :: thumbs up ::

PR #229, opened 2022-07-20
Cleanup operators

hlorus:
Splits out the different operators into separete files located in a subpackage

### DIFF
--- a/declarations.py
+++ b/declarations.py
@@ -1,8 +1,6 @@
 # Please keep this file in alphabetical order
 from enum import Enum
 
-from kiwisolver import Constraint
-
 
 class Gizmos(str, Enum):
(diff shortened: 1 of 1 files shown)


__init__.py
```diff
@@ -47,10 +47,11 @@ def register():
 
         global_data.registered = True
         logger.info("Solvespace available, fully registered modules")
-    except ModuleNotFoundError:
+    except ModuleNotFoundError as e:
         global_data.registered = False
         logger.warning(
-            "Solvespace module isn't available, only base modules registered"
+            "Solvespace module isn't available, only base modules registered\n"
+            + e
```
Cheaterman, line 54: Can't concatenate Exception to str? :-) (probably missing str(e) here)

PR #237, opened 2022-07-30
replace pi/tau constants with fractions of a turn

exegetor:
It's no secret that I'm a fan of TAU over PI, but better yet is to replace "magic numbers" with pre-defined constants to improve readability.
This especially benefits the less mathematically adept.

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -31,9 +31,9 @@ from .solver import solve_system, Solver
 from .functions import pol2cart, unique_attribute_setter
 from .declarations import Operators
 from .global_data import WpReq
+from .utilities.constants import FULL_TURN, HALF_TURN, QUARTER_TURN
 
 logger = logging.getLogger(__name__)
-
 
 class SlvsGenericEntity:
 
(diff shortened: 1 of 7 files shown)


class_defines.py
```diff
@@ -1270,8 +1269,8 @@ def to_bezier(
         if set_startpoint:
             startpoint.co = locations[0].to_3d()
 
-        n = 2 * math.pi / angle
-        q = (4 / 3) * math.tan(math.pi / (2 * n))
+        n = FULL_TURN / angle
+        q = (4 / 3) * math.tan(HALF_TURN / (2 * n))
```
exegetor, line 1273: I did a bit of algebra and it looks like it simplifies to  q= 4/3 tan(angle/4)
if so, I'll put it in the next PR

operators/add_angle.py
```diff
@@ -14,7 +16,7 @@ def invert_angle_getter(self):
 
 
 def invert_angle_setter(self, setting):
-    self["value"] = math.pi - self.value
+    self["value"] = QUARTER_TURN - self.value
```
hlorus, line 19: This should be half, right?

hlorus (commented):
IMO a great cleanup! I wouldn't put the definition into global_data.py tho. Maybe inside a constants.py in the root directory (probably that's declarations.py) or in /utilities/constants.py

operators/add_angle.py
exegetor: d'oh!

exegetor:
constants moved to new file, all imports updated, and the wrong value corrected.
should be OK now.
it's sleepy-time, so I shouldn't risk making any blunders now.  g'nite

class_defines.py
hlorus: I'm honestly not sure if such simplifications of formulas make a big performance improvement, i assume python optimizes some stuff anyway. But if it's not hurting readability it would be great of course.

Issue #40, opened 2022-04-26
[REQUEST] Selection modes


hlorus:
**Motivation / Existing Problem**
Native selection tools offer different modes, add the same to the addon's select tool.

**Description**
Add selection modes to allow adding/substracting/toggling selection. Idealy follow blender's default keymap, see: https://docs.blender.org/manual/en/latest/interface/selecting.html#selecting-modes

amrsoll:
I guess we should shoot for this selection menu, like when using the Blender selection tool?
![image](https://user-images.githubusercontent.com/24782867/183714492-bd9cffc5-d544-4532-84d0-0d573300a24d.png)

It is not shown when selecting the Solvespace selection tool (yet)
![image](https://user-images.githubusercontent.com/24782867/183714654-6602b37c-9ad6-4788-be92-ed5add8ed8b1.png)

hlorus:
> I guess we should shoot for this selection menu, like when using the Blender selection tool?
> ![image](https://user-images.githubusercontent.com/24782867/183714492-bd9cffc5-d544-4532-84d0-0d573300a24d.png)
> 
> It is not shown when selecting the Solvespace selection tool (yet)
> ![image](https://user-images.githubusercontent.com/24782867/183714654-6602b37c-9ad6-4788-be92-ed5add8ed8b1.png)
> 

Exactly! Although i'm not sure if the last two selection modes are really that useful...

hlorus:
This is now supported, see:
https://github.com/hlorus/CAD_Sketcher/commit/8b2a71dd6917addfaa891e4c80c43c22054a688d

What's remaining are the shortcut bindings. The problem here is that shift is currently used to select through gizmos.

hlorus:
The ctrl and shift modifiers are now also there, constraint selectability was moved to ctrl+shift.

PR #240, opened 2022-08-03
Selection modifiers

maciejgornicz:
_My first Open-source contribution :)_

Related to #40
Added features with shortcuts:

### select inverse `[CTRL + i]`
self-explained
### select extend `[CTRL + e]`
Works different than in [blender selecting-modes](https://docs.blender.org/manual/en/latest/interface/selecting.html#selecting-modes)
This one selects next entities in chain
### select extend all `[CTRL + SHIFT + e]`
Selects all entities in chain


operators/utilities.py
```diff
@@ -38,6 +38,43 @@ def select_all(context: Context):
...
+def select_extend(context: Context):
+    sketch = context.scene.sketcher.active_sketch
+    if sketch:
+        generator = sketch.sketch_entities(context)
+    else:
+        generator = entities_3d(context)
+    
+    to_select = []
+    for e in generator:
+        if e.is_line() or e.is_curve():
```
hlorus, line 60: You can probably use entity.is_segment() alone here

hlorus (changes requested):
Thanks for the PR, this really feels great!

I've experienced an issue when a circle is present:
  File "/home/yoga/Dropbox/blend/addons/CAD_Sketcher/operators/select.py", line 89, in execute
    while select_extend(context):
  File "/home/yoga/Dropbox/blend/addons/CAD_Sketcher/operators/utilities.py", line 65, in select_extend
    if e.p1.selected or e.p2.selected:
AttributeError: 'SlvsCircle' object has no attribute 'p1'
Error: Python: Traceback (most recent call last):
  File "/home/yoga/Dropbox/blend/addons/CAD_Sketcher/operators/select.py", line 89, in execute
    while select_extend(context):
  File "/home/yoga/Dropbox/blend/addons/CAD_Sketcher/operators/utilities.py", line 65, in select_extend
    if e.p1.selected or e.p2.selected:
AttributeError: 'SlvsCircle' object has no attribute 'p1'

Also it would be great to list the new shortcuts in the documentation, there's already some hints for the selection tool in docs/tools.md.

Hope to see you around :)

operators/utilities.py
maciejgornicz: done

operators/utilities.py
```diff
@@ -38,6 +38,52 @@ def select_all(context: Context):
...
+    if sketch:
+        generator = sketch.sketch_entities(context)
+    else:
+        generator = entities_3d(context)
+    
+    to_select = []
+    for e in generator:
+        if e.is_segment():
+            if e.selected:
+                if hasattr(e, 'p1'):
```
hlorus, line 62: Probably easier to just append e.connection_points()

operators/utilities.py
```diff
@@ -38,6 +38,52 @@ def select_all(context: Context):
...
+                    to_select.append(e.p1)
+                if hasattr(e, 'p2'):
+                    to_select.append(e.p2)
+                continue
+            if (hasattr(e,'p1') and e.p1.selected) or (hasattr(e, 'p2') and e.p2.selected):
+                to_select.append(e)    
+                continue
+    
+    for coindident in context.scene.sketcher.constraints.coincident:
```
hlorus, line 71: This loop gets quite complicated, better to early exit the loop by checking the inversed if statements and continue if True, e.g.:
`
for c in ...coincident:
  if not entity.selected:
    continue
`
And the same for the inner loop.

hlorus (changes requested):
Sorry to bother you again..

.gitignore
```diff
@@ -133,3 +133,6 @@ notes.md
 
 # Utilities
 utils/*
+
+#vscode settings
+.vscode/*
```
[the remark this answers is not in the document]
amrsoll: It depends on how strictly you want to enforce how to do linting etc.
In general it is good to keep track of workspace-specific settings. For developer preferences and add-ons, the global settings exist (and will not be stored in the repository itself)

operators/utilities.py
```diff
@@ -38,6 +38,52 @@ def select_all(context: Context):
...
+                    to_select.append(entity_to_select)
+                continue    
+            
+    is_something_to_select = False
+    for entity in to_select:
+        if not entity.selected:
+            is_something_to_select = True
+            entity.selected = True
+    
+    return is_something_to_select
```
amrsoll, line 84: Slight simplification:
```py
is_something_to_select = not all(e.selected for e in to_select)
for entity in to_select:
    entity.selected = True
```

amrsoll: Secondly, why would you return whether something has been selected or not?

operators/utilities.py
amrsoll: although it would be best if we could have a garantee that all elements have that function (probably should be implemented for the abstract class SlvsGenericEntity)

amrsoll: but yes, definitely better to use `connection_points()` as `p1` and `p2` are way too specific. I would recommend using `not isinstance(e, class_defines.point) and isinstance(e, SlvsGenericEntity)`

operators/utilities.py
```diff
@@ -38,6 +38,52 @@ def select_all(context: Context):
...
+    to_select = []
+    for e in generator:
+        if e.is_segment():
+            if e.selected:
+                if hasattr(e, 'p1'):
+                    to_select.append(e.p1)
+                if hasattr(e, 'p2'):
+                    to_select.append(e.p2)
+                continue
+            if (hasattr(e,'p1') and e.p1.selected) or (hasattr(e, 'p2') and e.p2.selected):
```
amrsoll, line 67: This should be an `elif` so you avoid adding elements that are already selected.

operators/utilities.py
```diff
@@ -38,6 +38,52 @@ def select_all(context: Context):
...
+                to_select.append(e)    
+                continue
+    
+    for coindident in context.scene.sketcher.constraints.coincident:
+        for entity in coindident.entities():
+            if entity.selected:
+                for entity_to_select in coindident.entities():
+                    to_select.append(entity_to_select)
+                continue
```
amrsoll, line 76:
```suggestion
    for coindident in context.scene.sketcher.constraints.coincident:
        if any(entity.slected for entity in coindident.entities()):
            to_select.extend(coindident.entities())
```

operators/utilities.py
amrsoll: if you this return value does not matter, I would suggest to remove the `to_select` variable and simply write `entity.selected = True` where applicable.

amrsoll (changes requested):
All in all, cool feature to have, good work :smile:

operators/utilities.py
amrsoll: typo coin**d**ident

operators/utilities.py
maciejgornicz: i'm looping until select_extend return False (in select.py). This is needed for extend sleection with all in chain.

operators/utilities.py
maciejgornicz: @amrsoll @hlorus  Maybe it is good idea to create mixin class like `SegmentEntity` insted/additional of `is_segment` method. We could add `connection_points()` and additionally create `coincident_points()` method to it?

.gitignore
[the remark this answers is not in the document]
maciejgornicz: removed 👍

operators/utilities.py
maciejgornicz: Done 👍 Thanks for pointing

operators/utilities.py
maciejgornicz: Done 👍

operators/select.py
```diff
@@ -2,10 +2,13 @@
...
 from bpy.utils import register_classes_factory
 
-from .utilities import select_all, deselect_all
+from .utilities import select_all, deselect_all, select_extend, select_invert
 from .. import global_data
 from ..declarations import Operators
 from ..utilities.highlighting import HighlightElement
+import logging
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
maciejgornicz: I've done logging in `utilities.py` so I've removed from here

operators/utilities.py
maciejgornicz: i Simplified this as @amrsoll sugested 👍

operators/utilities.py
amrsoll: I have a feeling that these classes might change in the not-so-distant future and adding the extra mixin class isn't justifiable.

maciejgornicz: copy that

hlorus (approved):
Just note some small details. IMO this can be merged if @amrsoll is also fine with the current state.
Thanks alot for the work all!

operators/utilities.py
```diff
@@ -36,8 +38,49 @@ def select_all(context: Context):
...
+    if sketch:
+        logger.debug(f"Extending chain selection of sketcher entities in sketch : {sketch.name} (slvs_index: {sketch.slvs_index})")
+        generator = sketch.sketch_entities(context)
+    else:
+        logger.debug(f"Extending chain selection of sketcher entities")
+        generator = entities_3d(context)
+    
+    to_select = []
+    for e in generator:
+        if not isinstance(e, point) and isinstance(e, SlvsGenericEntity):
```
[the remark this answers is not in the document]
maciejgornicz: done 👍

operators/select.py
```diff
@@ -54,5 +55,43 @@ def execute(self, context: Context):
...
+
+class View3D_OT_slvs_select_extend_all(Operator):
+    """Select neighbour entities"""
+
+    bl_idname = Operators.SelectExtendAll
+    bl_label = "Select neighbour entities"
+
+    def execute(self, context: Context):
+        while select_extend(context):
+            context.area.tag_redraw()
```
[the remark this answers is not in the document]
maciejgornicz: done 👍

operators/select.py
```diff
@@ -2,10 +2,11 @@
 from bpy.props import IntProperty, BoolProperty
 from bpy.utils import register_classes_factory
 
-from .utilities import select_all, deselect_all
+from .utilities import select_all, deselect_all, select_extend, select_invert
 from .. import global_data
 from ..declarations import Operators
 from ..utilities.highlighting import HighlightElement
+import logging
```
amrsoll, line 9: not used?

operators/utilities.py
```diff
@@ -7,7 +7,7 @@
 
 from .. import global_data
 from ..declarations import GizmoGroups, WorkSpaceTools
-from ..class_defines import SlvsGenericEntity
+from ..class_defines import SlvsGenericEntity, point
```
amrsoll, line 10: now that we changed the `if` statement, the `point` import isn't useful :)

amrsoll (changes requested):
Minor changes because of unused imports :)

operators/select.py
maciejgornicz: I'll pay more attention :)

operators/utilities.py
maciejgornicz: done

amrsoll (approved):
All good for squash merge :) thank you for the contribution!

hlorus:
🎉🎊
Alright then, i'm gonna merge this. Thanks @maciejgornicz for the patience!

PR #241, opened 2022-08-04
make consistant order of member variables, spacings, and parenthesis;…

exegetor:
Since the apostropy is use both for contractions and possessives,  the famously insane English language has had to face the dilemma presented by the word "it".
An arbitrary decision was made to go with "its" for possessive and "it's" for "it is"

The other notable change here is to make sure all the bl_* variables are listed first, with bl_options always the last of those.

A couple of line spacing and parenthesis changes aren't worth mentioning.


hlorus:
LGTM, thanks alot!

PR #246, opened 2022-08-10
Add missing solve return values

abergmeier:
According to C headers there are values 4 and 5 possible.

### DIFF
--- a/global_data.py
+++ b/global_data.py
@@ -32,33 +32,53 @@ class WpReq(Enum):
 
 
 solver_state_items = [
-    ("OKAY", "Okay", "Successfully solved sketch", "CHECKMARK", 0),
+    (
+        "OKAY",
+        "Okay",
+        "Successfully solved sketch",
+        "CHECKMARK",
+        0,  # SLVS_RESULT_OKAY
+    ),
     (
         "INCONSISTENT",
         "Inconsistent",
         "Cannot solve sketch because of inconsistent constraints",
         "ERROR",
-        1,
+        1,  # SLVS_RESULT_INCONSISTENT
     ),
     (
         "DIDNT_CONVERGE",
         "Didnt Converge",
         "Cannot solve sketch, system didn't converge",
         "ERROR",
-        2,
+        2,  # SLVS_RESULT_DIDNT_CONVERGE
     ),
     (
         "TOO_MANY_UNKNOWNS",
         "Too Many Unknowns",
         "Cannot solve sketch because of too many unknowns",
         "ERROR",
-        3,
+        3,  # SLVS_RESULT_TOO_MANY_UNKNOWNS
+    ),
+    (
+        "INIT_ERROR",
+        "Initialize Error",
+        "Initializing solver failed",
+        "ERROR",
+        4,  # SLVS_RESULT_INIT_ERROR
+    ),
+    (
+        "REDUNDANT_OK",
+        "Redundant Ok",
+        "Constraints seem redundand",
+        "CHECKMARK",
+        5,  # SLVS_RESULT_REDUNDANT_OK
     ),
     (
         "UNKNOWN_FAILURE",
         "Unknown Failure",
         "Cannot solve sketch because of unknown failure",
         "ERROR",
-        4,
+        6,
     ),
 ]
--- a/solver.py
+++ b/solver.py
@@ -211,18 +211,16 @@ class Solver:
                 findFreeParams=False,
             )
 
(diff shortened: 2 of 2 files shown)


abergmeier:
This currently is missing two things:
1. it does not force re-solving. Maybe this should be done for every new release/code anyway?
2. It does not show which constraints are redundant.

hlorus:
Nice fix! I think the documentation was wrong as the 5th error was never explained. Now that all the possible errors are implemented the "Solver returned unknown error" could probably be removed.

hlorus:
> This currently is missing two things:
> 1. it does not force re-solving. Maybe this should be done for every new release/code anyway?
> 2. It does not show which constraints are redundant. 

I don't really get what you mean with #1 here.

#2 Afaik solvespace doesn't have support to report which exact constraints are redundant/overconstrained, same with degree of freedom.

abergmeier:
> I don't really get what you mean here.

Since previously the code did clamp e.g. 5 to 4 if you open a blender file with the new code it shows thus the error message for 4 - which is actually wrong.

abergmeier:
> Afaik solvespace doesn't have support to report which exact constraints are redundant/overconstrained, same with degree of freedom.

I am not sure. Looking at the C code where they do further redundant handling it seems like they do some more processing. Will try to look into this the next days some more.

hlorus:
Ahh i see, IMO this is really just a minor issue which isn't a big deal in pre 1.0 versions. Otherwis it could be adjusted in versioning.py.

abergmeier:
So since `SolveSpaceUI::SolveGroupAndReport` runs solving twice when it returns redundancies - maybe we should adopt the same mechanism?

hlorus:
Just re-try solving again whenever the solver fails or just with a specific return value of the solver?

hlorus:
@abergmeier This PR is IMO in a good state but is still marked as WIP... I'd say not re-solving old files when opening isn't a big deal as we haven't reached 1.0 yet and we also have a manual update button.
Additional changes like solving twice for redundant error is also better kept in a separate PR.

solver.py
```diff
@@ -211,18 +211,16 @@ def solve(self, report=True):
...
-                retval = 4
+
             self.result = bpyEnum(solver_state_items, index=retval)
 
             if report and sketch:
                 sketch.solver_state = self.result.index
                 sketch.dof = self.solvesys.Dof
 
-            if retval != 0:
+            if retval != 0 and retval != 5:
```
amrsoll, line 223: Why are redundant constraints not an error state?

abergmeier: Two reasons basically:
1. That is the way the C++ library used actually handles these things in their upper layers (note that CAD Sketcher uses lower layers).
2. From what I could gather there seems to be no (easy) way of detecting these redundant constraints.

amrsoll (approved):

amrsoll:
Seems to work brilliantly, except perhaps for the "redundant" + Ok warning message.

image

solver.py
amrsoll: Now that I tested it, it is really nice to only have it as a warning and not an error, as the solver seems to ignore duplicates/redundant constraints :)

global_data.py
```diff
@@ -32,33 +32,53 @@ class WpReq(Enum):
...
+    ),
+    (
+        "REDUNDANT_OK",
+        "Redundant Ok",
+        "Constraints seem redundand",
+        "CHECKMARK",
+        5,  # SLVS_RESULT_REDUNDANT_OK
     ),
     (
         "UNKNOWN_FAILURE",
```
hlorus, line 78: Do we still need the unknown failiure? I've used this because the solver was sometimes returning an undocumented value. Since you've looked at the source, is there a case where the solver returns the value 6?

hlorus:
This seems to work great, just some minor notes:
- It's easy to overlook the text change to "Redundant OK", maybe we could use a different icon if there's just a note, maybe the "INFO" icon?
- The constraints that are reported by the solver will be marked with "Failed" which might be a bit misleading.

global_data.py
abergmeier: Need them - strickly no. Then again if in the future the library decides to provide other return states you can pretty fast notice because users will report "Unknown failure".

amrsoll:
> The constraints that are reported by the solver will be marked with "Failed" which might be a bit misleading.

I think it is OK given there is no other way to know which constraints are redundant atm.

global_data.py
hlorus: Makes sense...

PR #247, opened 2022-08-10
Add some more type hints.

abergmeier:


class_defines.py
```diff
@@ -737,7 +737,7 @@ def hide_sketch(self, context):
     target_object: PointerProperty(type=bpy.types.Object)
     curve_resolution: IntProperty(name="Mesh Curve Resolution", default=12, min=1, soft_max=25)
 
-    def dependencies(self):
+    def dependencies(self) -> List[SlvsWorkplane]:
```
hlorus, line 740: Probably List[SlvsGenericEntity] makes sense here has the method is implemented on all entity types.

hlorus (changes requested):

class_defines.py
abergmeier: Done

hlorus (approved):

hlorus:
LGTM

PR #265, opened 2022-08-22
Default constraint names use the str() function.

### LINKED ISSUE #263: Added a radius dimension to a circle, but the entity type shows as 'SlvsDiameter'
### Contact Details

paynterf@gmail.com

### Description

I added a circle entity to a new sketch and selected ALT+R to add a radius dimension.  I got the normal radius depiction and was able to enter a radius, but the dimensional entity shows as 'SlvsDiameter', not 'SlvsRadius'.  With just this information I can't tell whether the circle has a radius of 15mm or a diameter of 15mm.

Also, I don't seem to be able to select the dimension arrow so I can move it around like I could with 0.24

### Addon Version

0.25

### Blender Version

3.2.1

Windows

amrsoll:
This reverts the default naming back to the previously used `label`.

In the case of the diameter constraint, it will return either "Radius" or "Diameter"
depending on the setting.

If you edit the name of the Diameter constraint, it replaces the default naming.

Fixes #263 

Note: When you switch the constraint between Diameter <-> Radius, it changes the constraint name, unless it is user defined.

image

Let me know If I should do the same for the `SlvsGenericEntity`s.

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -2179,11 +2179,9 @@ ENTITY_PROP_NAMES = ("entity1", "entity2", "entity3", "entity4")
 
 
 class GenericConstraint:
+
     def _name_getter(self):
-        name = self.get("name")
-        if not name:
-            return type(self).__name__
-        return name
+        return self.get("name", str(self))
 
     def _name_setter(self, new_name):
         self["name"] = new_name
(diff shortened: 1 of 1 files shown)


hlorus:
LGTM!

Issue #261, opened 2022-08-19
Angel Constraint Bug - $50


hlorus:

amrsoll:
Uncaught exception happens when solving the system

![uncaught exception](https://user-images.githubusercontent.com/24782867/186008710-ffabfcf3-1487-49c9-bb4a-a32ae6e750b3.png)

Pointing to [this particular line](https://github.com/realthunder/solvespace/blob/a2b33c81a82cb158db2fa8748bbd161816b23abc/src/entity.cpp#L528)

amrsoll:
Even more minimal example using 2 lines, vertical or horizontal constraint + vertical / horizontal Distance constraint

![image](https://user-images.githubusercontent.com/24782867/186137032-b89bb9c0-15ad-4a56-93a3-1a36a13ffe20.png)

hlorus:
> Compared to the upstream solvesystem repo, the one we use with the python bindings is 1100 commits behind. I think the realthunder/solvespace repo desperatly needs an update from upstream.
> 

Absolutely agree with that, we should probably reach out to realthunder.


> It might be worth trying this version instead: https://github.com/KmolYuan/solvespace
>  https://pypi.org/project/python-solvespace/

That's actually what i used to start with but i've decided to switch to the version that is used now back because it didn't seem to be maintained iirc.


> But I think @hlorus knows best about the different python ports of solvespace.

I don't really know alot about the differences. There's just no available python package of the official solvespace solver as far as i know.

amrsoll:
> it didn't seem to be maintained iirc.

The [https://github.com/KmolYuan/python-solvespace](https://github.com/KmolYuan/python-solvespace ) is archived, but he also has https://github.com/KmolYuan/solvespace which is maintained and a lot closer to upstream. I'm trying to use it atm, but there are quite a few differences (obvi)

therealjacobbrauer:
> Uncaught exception happens when solving the system
> 
> ![uncaught exception](https://user-images.githubusercontent.com/24782867/186008710-ffabfcf3-1487-49c9-bb4a-a32ae6e750b3.png)
> 
> Pointing to [this particular line](https://github.com/realthunder/solvespace/blob/a2b33c81a82cb158db2fa8748bbd161816b23abc/src/entity.cpp#L528)
> 
> Compared to the upstream solvesystem repo, the one we use with the python bindings is 1100 commits behind. I think the realthunder/solvespace repo desperatly needs an update from upstream.
> 
[...]

PR #268, opened 2022-08-23
Refactor for python_solvespace API.

amrsoll:
## Why?

The PR is currently a draft to feel the temperature of the water and talk about it.

The python_solvespace repo is much closer to the upstream solvespace repo. This means that some bugs might already be solved there, such as https://github.com/hlorus/CAD_Sketcher/issues/261 (not confirmed if this can fix it)

A side effect of this refactor is that a lot of abstraction is already done by the python wrapper :smile: and saves a lot of lines.

## History

@hlorus used to base CAD sketcher on [an other solvespace repo](https://github.com/KmolYuan/python-solvespace) . As it got archived it seemed to be abandonned, but an [other fork](https://github.com/KmolYuan/solvespace) was maintained by the same user.

## TODO

- [ ] Fix updating entity properties post-solving
- [ ] Fix dragging entities
- [ ] Work out how to use the groups with this API (I pretty much deleted everything related to groups)
- [ ] Update docs
- [ ] Fix vertical / horizontal constraints
- [ ] A lot more

### DIFF
--- a/class_defines.py
+++ b/class_defines.py
@@ -123,7 +123,7 @@ class SlvsGenericEntity:
 
     @property
     def py_data(self):
-        return global_data.entities[self.slvs_index]
+        return global_data.entities.get(self.slvs_index, None)
 
     @py_data.setter
     def py_data(self, handle):
@@ -286,9 +286,14 @@ class SlvsGenericEntity:
         gpu.shader.unbind()
         self.restore_opengl_defaults()
 
-    def create_slvs_data(self, solvesys):
-        """Create a solvespace entity from parameters"""
-        raise NotImplementedError
+    def create_slvs_data(self, solvesys, *args, **kwargs):
+        """Create a solvespace entity from parameters
+        Also makes sure that all the dependencies have their data generated.
+        """
+        for e in self.dependencies():
+            if isinstance(e, SlvsGenericEntity) and not getattr(e, "py_data", None):
+                e._create_slvs_data(solvesys)
+        return self._create_slvs_data(solvesys, *args, **kwargs)
 
     def update_from_slvs(self, solvesys):
         """Update parameters from the solvespace entity"""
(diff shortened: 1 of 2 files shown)


exegetor:
You said "same user" but I don't see any sign of it on realthunder's page.  Can you link to the python_solvespace repo you're using?

amrsoll:
done :)

hlorus:
Definitely intresting to explore this direction. My main question here is if we would have to drop support for the current python port "py_slvs".

Some people also pointed completely different solvers out. Maybe we could start thinking about having it somewhat generic to allow plugging in any solver we want in the future.

amrsoll:
> My main question here is if we would have to drop support for the current python port "py_slvs".

There doesn't immediately seem like there is any advantage of using py_slvs over python_solvespace (god I hate when python packages put py/python in their name :joy: )

> Maybe we could start thinking about having it somewhat generic to allow plugging in any solver we want in the future.

Oh the nightmare :joy: given how differently some solvers and their API might work (imagine something acting as weirdly as `bpy`) it will be integration hell to have support for multiple solvers, and I don't think they can be easily abstracted.

hlorus:
>> Oh the nightmare :joy: given how differently some solvers and their API might work (imagine something acting as weirdly as `bpy`) it will be integration hell to have support for multiple solvers, and I don't think they can be easily abstracted.

Hmm yeah, might not be worth the hassle to support multiple solvers.

hlorus:
Interestingly i'm already failing at installing the solver...

hlorus:
This will likely need alot of testing so i would recommend making this an official branch, getting it to a testable stage and then collect feedback by testers.

amrsoll:
Well i am happy not to be the only one to want this tested out, see if it works better :)

amrsoll:
Closing this PR as a duplicate of #320

PR #269, opened 2022-08-24
Fix bevel issue when using arcs.

### LINKED ISSUE #267: Bevel Tool: Correct center point of the arc - $50
When creating a bevel the proper placement of the arc has to be found. Currently this is done by first offseting the neighboring segments(line, arc) by the radius and then searching for intersections between them. The problem here is that there will be multiple intersections and its not trivial to get the correct intersection which leads to bad behavior:
ezgif com-gif-maker

To solve this it might be necessary to implement specific equation for the different scenarios e.g. line-line / line-arc /arc-arc.

amrsoll:
Fixes #267 

All the hard work was done beforehand, just had to tweak some knobs here and there :joy:

### DIFF
--- a/operators/bevel.py
+++ b/operators/bevel.py
@@ -137,19 +137,24 @@ class View3D_OT_slvs_bevel(Operator, Operator2d):
         l1, l2 = connected
         self.connected = connected
 
-        intersections = _get_intersections(
-            *_get_offset_elements(l1, radius),
-            *_get_offset_elements(l2, radius),
+        # If more than 1 intersection point, then sort them so we prioritise 
+        # the closest ones to the selected point.
+        #   (Can happen with intersecting arcs)
+        intersections = sorted(
+            _get_intersections(
+                *_get_offset_elements(l1, radius),
+                *_get_offset_elements(l2, radius),
+            ),
+            key=lambda i: (i - self.p1.co).length
         )
 
         coords = None
-        for i in intersections:
-
-            if hasattr(l1, "is_inside") and not l1.is_inside(i):
+        for intersection in intersections:
+            if hasattr(l1, "is_inside") and not l1.is_inside(intersection):
                 continue
-            if hasattr(l2, "is_inside") and not l2.is_inside(i):
+            if hasattr(l2, "is_inside") and not l2.is_inside(intersection):
                 continue
-            coords = i
+            coords = intersection
             break
 
         if not coords:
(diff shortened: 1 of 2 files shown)


exegetor:
works great with (double) intersecting arcs but then I found this

https://user-images.githubusercontent.com/10360161/186289585-f5724618-9be8-4188-ab25-e8e2d53474c0.mp4

amrsoll:
I made a second force push 20 min ago that should have fixed what you have tested. I am not able to reproduce it though.

exegetor:
very well, I cannot get replicate previous video, but while trying, I found this

https://user-images.githubusercontent.com/10360161/186302951-27560ace-ed7c-406a-917f-42e66d968248.mp4


Debug client attached.
CAD_Sketcher-main:{INFO}: Logging into: C:\Users\shann\AppData\Local\Temp\CAD_Sketcher-main.log
Read blend: C:\Users\shann\Documents\dev\blender\bevel_arc.blend
Python: Traceback (most recent call last):
  File "C:\Users\shann\AppData\Roaming\Blender Foundation\Blender\3.1\scripts\addons\CAD_Sketcher-main\utilities\highlighting.py", line 81, in invoke
    return self.execute(context)
  File "C:\Users\shann\AppData\Roaming\Blender Foundation\Blender\3.1\scripts\addons\CAD_Sketcher-main\operators\delete_entity.py", line 100, in execute
    self.main(context, index, self)
  File "C:\Users\shann\AppData\Roaming\Blender Foundation\Blender\3.1\scripts\addons\CAD_Sketcher-main\operators\delete_entity.py", line 53, in main
    operator.delete(entities.get(i), context)
  File "C:\Users\shann\AppData\Roaming\Blender Foundation\Blender\3.1\scripts\addons\CAD_Sketcher-main\operators\delete_entity.py", line 87, in delete
    logger.debug("Delete: {}".format(data_coll[i]))
IndexError: bpy_prop_collection[index]: index 1 out of range, size 1

location: <unknown location>:-1

operators/bevel.py
```diff
@@ -137,19 +137,24 @@ def main(self, context):
         l1, l2 = connected
         self.connected = connected
 
-        intersections = _get_intersections(
-            *_get_offset_elements(l1, radius),
-            *_get_offset_elements(l2, radius),
+        # If more than 1 intersection point, then sort them so we prioritise 
+        # the closest ones to the selected point.
```
hlorus, line 141: I'm not sure about that. The closest one doesn't have to be the correct one. It will probably often be the case but not when creating a big bevel on a symmetrical setup like described in #267.

amrsoll: I feel like that isn't possible to create that case with straight lines and arcs.

I could relatively easily create a solution for any arc-line combination for bevel where you can come up with the path along which the centerpoint moves (which is an arc I think) and use the radius with the function [`distance_along_segment`](https://github.com/hlorus/CAD_Sketcher/blob/main/class_defines.py#L1531) to get the centerpoint.

That allows to create a bevel along the whole length of both entities too, instead of half of them in some scenario.

However, it ain't really a general solution either, and a general solution sounds like it would be done on the library side, either slvs or blender (bpy) itself, but would require native blender objects.

So unless we want to maintain some very complex code that already exists in other libraries, i think this solution or the other one I suggested should be considered.

amrsoll:
In the end I don't really care about the bounty, I'm just happy to aupport the project this way :) I'm okay not receiving the bounty if that is what's holding back the PR.

I found out that even the second method of generating a path for the center point along an arc is more complicated (annoying) than I thought, and I really don't think it is worth putting the time into that if it's possible to (re)use a more complete/generalised bevel tool down the line.

hlorus:
I will check the PR next week to see what the remaining cornercases are.
It is indeed quite complicated and in theory already available in native blender but it's also a much asked for feature...

hlorus:
I gave this a try and IMO that's a great improvement. In the case of two entities that form a closed shape it now simply fails when trying to add a bevel that goes further than the middle of the segments which is a minor limitation.

PR #272, opened 2022-08-24
Create "constructor" constraints.

### LINKED ISSUE #258: Reference Dimensions
### Discussed in https://github.com/hlorus/CAD_Sketcher/discussions/84

<div type='discussions-op-text'>

<sup>Originally posted by **TheMakerTales** April 30, 2022</sup>
A "construction" mode for dimensional constraints. Sometimes you want to know how long or round a part of a sketch is but not actually contain it.</div>

amrsoll:
Fixes #258

20220914_12h50m35s_grim
20220914_12h47m27s_grim

## TODO

- [x] Update constraint value from the attached entity instead of the solver .. hmmm
- [x] Change look
- [x] Make dimension read only when `is_reference` is True.
- [x] Possibly create `class DimensionalConstraint` in `base_constraint.py`
- [x] Fix bug where color is not proper during mouse hover right after switching reference mode
- [x] Revert change where reference constraints are filtered out of the constraints menu

### DIFF
--- a/model/base_constraint.py
+++ b/model/base_constraint.py
@@ -5,7 +5,7 @@ from bpy.props import StringProperty, BoolProperty
 from bpy.types import UILayout
 
 from .. import functions
-from ..solver import solve_system
+from ..solver import solve_system, Solver
 from ..global_data import WpReq
 from ..utilities import preferences
 from ..declarations import Operators
@@ -28,6 +28,7 @@ class GenericConstraint:
     is_reference: BoolProperty(name="Construction")
     signature = ()
     props = ()
+    dirty: BoolProperty(name="Needs updating", default=False)
 
     def needs_wp(args):
         return WpReq.OPTIONAL
--- a/model/base_constraint.py
+++ b/model/base_constraint.py
@@ -201,10 +201,15 @@ class DimensionalConstraint(GenericConstraint):
         # self.value, self.setting = self.init_props()
         self.value, _ = self.init_props()
 
+    def refresh_constraint(self, context: Context=None):
+        self.assign_init_props()
+        # Refresh the gizmos as we are changing the colors.
+        functions.refresh(context)
+
     is_reference: BoolProperty(
         name="Only measure",
         default=False,
-        update=assign_init_props,
+        update=refresh_constraint,
     )
 
     def init_props(self):
--- a/model/base_constraint.py
+++ b/model/base_constraint.py
@@ -2,6 +2,7 @@ import logging
 from typing import List
 
 from bpy.props import StringProperty, BoolProperty
+import bpy.types
 from bpy.types import UILayout
 
 from .. import functions
@@ -25,7 +26,6 @@ class GenericConstraint:
     name: StringProperty(name="Name", get=_name_getter, set=_name_setter)
     failed: BoolProperty(name="Failed")
     visible: BoolProperty(name="Visible", default=True, update=functions.update_cb)
-    is_reference: BoolProperty(name="Construction")
     signature = ()
     props = ()
     dirty: BoolProperty(name="Needs updating", default=False)
@@ -152,7 +152,6 @@ class GenericConstraint:
         # General props
         layout.separator()
         layout.prop(self, "visible")
-        layout.prop(self, "is_reference")
 
         # Specific props
         layout.separator()
@@ -176,8 +175,33 @@ class GenericConstraint:
     def placements(self):
         """Return the entities where the constraint should be displayed"""
         return []
-    
+
+    def create_slvs_data(self, solvesys: Solver, **kwargs):
+        raise NotImplementedError()
+
+    def py_data(self, solvesys: Solver, **kwargs):
+        return self.create_slvs_data(solvesys, **kwargs)
+
+class DimensionalConstraint(GenericConstraint):
+
+    def make_readonly(self, context: bpy.types.Context):
+        print("here we are")
+        return None
+
+    is_reference: BoolProperty(
+        name="Only measure",
+        default=False,
+    )
+
     def py_data(self, solvesys: Solver, **kwargs):
         if self.is_reference:
             return []
         return self.create_slvs_data(solvesys, **kwargs)
+
+    def draw_props(self, layout: UILayout):
+        sub = GenericConstraint.draw_props(self, layout)
+        sub.prop(self, "is_reference")
+        return sub.column()
+    
+    def is_active(self, *args, **kwargs):
+        return GenericConstraint.is_active(self, *args, **kwargs)
--- a/model/base_constraint.py
+++ b/model/base_constraint.py
@@ -25,6 +25,7 @@ class GenericConstraint:
     name: StringProperty(name="Name", get=_name_getter, set=_name_setter)
     failed: BoolProperty(name="Failed")
     visible: BoolProperty(name="Visible", default=True, update=functions.update_cb)
+    is_reference: BoolProperty(name="Construction")
     signature = ()
     props = ()
 
@@ -150,6 +151,7 @@ class GenericConstraint:
         # General props
         layout.separator()
         layout.prop(self, "visible")
+        layout.prop(self, "is_reference")
 
         # Specific props
         layout.separator()
@@ -173,3 +175,8 @@ class GenericConstraint:
     def placements(self):
         """Return the entities where the constraint should be displayed"""
         return []
+    
+    def py_data(self, solvesys: Solver, **kwargs):
+        if self.is_reference:
+            return []
+        return self.create_slvs_data(solvesys, **kwargs)
--- a/model/base_constraint.py
+++ b/model/base_constraint.py
@@ -188,6 +188,13 @@ class DimensionalConstraint(GenericConstraint):
     setting: BoolProperty
 
     def _set_value(self, val: float):
+        # NOTE: function signature _set_value(self, val: float, force=False)
+        #       will fail when bpy tries to register the value property.
+        #       See `_set_value_force()`
+        if not self.is_reference:
+            self._set_value_force(val)
+    
+    def _set_value_force(self, val: float):
         self["value"] = val
 
     def _get_value(self):
@@ -200,9 +207,10 @@ class DimensionalConstraint(GenericConstraint):
 
     def assign_init_props(self, context: Context = None):
         # self.value, self.setting = self.init_props()
-        self.value, _ = self.init_props()
+        _value, _ = self.init_props()
+        self._set_value_force(_value)
 
-    def refresh_constraint(self, context: Context = None):
+    def on_reference_checked(self, context: Context = None):
         self.assign_init_props()
         # Refresh the gizmos as we are changing the colors.
         functions.refresh(context)
@@ -210,7 +218,7 @@ class DimensionalConstraint(GenericConstraint):
     is_reference: BoolProperty(
         name="Only measure",
         default=False,
-        update=refresh_constraint,
+        update=on_reference_checked,
     )
 
     def init_props(self):
--- a/model/base_constraint.py
+++ b/model/base_constraint.py
@@ -181,6 +181,7 @@ class GenericConstraint:
     def py_data(self, solvesys: Solver, **kwargs):
         return self.create_slvs_data(solvesys, **kwargs)
 
+
 class DimensionalConstraint(GenericConstraint):
 
     value: Property
@@ -197,11 +198,11 @@ class DimensionalConstraint(GenericConstraint):
             self.assign_init_props()
         return self["value"]
 
-    def assign_init_props(self, context: Context=None):
+    def assign_init_props(self, context: Context = None):
         # self.value, self.setting = self.init_props()
         self.value, _ = self.init_props()
 
-    def refresh_constraint(self, context: Context=None):
+    def refresh_constraint(self, context: Context = None):
         self.assign_init_props()
         # Refresh the gizmos as we are changing the colors.
         functions.refresh(context)
@@ -231,11 +232,11 @@ class DimensionalConstraint(GenericConstraint):
             # Could not find a way to have the property "readonly",
             # so we disable user input instead
             col.prop(self, "value")
-            col.enabled=not self.is_reference
+            col.enabled = not self.is_reference
         if hasattr(self, "setting"):
             row = sub.row()
             row.prop(self, "setting")
         return sub
-    
+
     def is_active(self, *args, **kwargs):
         return GenericConstraint.is_active(self, *args, **kwargs)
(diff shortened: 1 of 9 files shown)


solver.py
```diff
@@ -121,10 +121,11 @@ def _get_msg_entities():
...
 
-            indices = c.create_slvs_data(self.solvesys, group=group)
-            self._store_constraint_indices(
-                c, indices if isinstance(indices, Iterable) else (indices,)
-            )
+            indices = c.py_data(self.solvesys, group=group)
+            if indices:
+                self._store_constraint_indices(
+                    c, indices if isinstance(indices, Iterable) else (indices,)
+                )
```
amrsoll, line 128: Not sure if the extra `if indices` is required if we return an empty list.

hlorus:
> * Update constraint value from the attached entity instead of the solver .. hmmm
Note: init_props already computes the constraints value and setting from it's entities. Would be great to reuse/split-out that code.

model/diameter.py
```diff
@@ -108,6 +123,10 @@ def update_draw_offset(self, pos, ui_scale):
 
     def draw_props(self, layout):
         sub = super().draw_props(layout)
+        # self.bl_rna.properties.get('value').is_readonly = True
+        #   Traceback (most recent call last):
+        #     File "<string>", line 1, in <module>
+        #   AttributeError: bpy_struct: attribute "is_readonly" from "FloatProperty" is read-only
```
amrsoll, line 129: @hlorus I have been unsuccessful at editing this attribute after the creation of the property.

Do you have knowledge on how to do this? I have not found any helpful resources for this specific case.

hlorus: I'm afraid property definitionns are not editable after registration. I would suggest to define our own getters/setters for the value properties and then just not set the value if it's a reference dimension.
https://docs.blender.org/api/current/bpy.props.html#getter-setter-example

hlorus:
Just found out about this:

>>> bpy.ops.wm.properties_edit(
properties_edit()
bpy.ops.wm.properties_edit(data_path="", property_name="", property_type='FLOAT', is_overridable_library=False, description="", use_soft_limits=False, array_length=3, default_int=(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0), min_int=-10000, max_int=10000, soft_min_int=-10000, soft_max_int=10000, step_int=1, default_float=(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0), min_float=-10000, max_float=-10000, soft_min_float=-10000, soft_max_float=-10000, precision=3, step_float=0.1, subtype='NONE', default_string="", eval_string="")
Change a custom property's type, or adjust how it is displayed in the interface

model/diameter.py
amrsoll: I was initially going to do that, but it is always preferable to sanitise user input rather than handling issues down the line ^^
Thankfully you seem to have found the way :smile:

amrsoll:
Well I found out that you cannot make Property values "readonly" for the user on their own, but that you have to declare it in any menu they appear in... or?? 

Anyway, the functionality is there, I am not super happy with the code structure I went with, I'll have to revise it perhaps..

hlorus:
Gave it a quick test and it's just awesome when you make a constraint depend on the value of a reference dimension and interactivley drag stuff around!
Is it intentional that reference dimensions don't show up in the constraint browser? IMO that's important to allow adding drivers which will probably be the biggest use-case for this.

amrsoll:
> Is it intentional that reference dimensions don't show up in the constraint browser? 

Yep, didn't think about the driver situation :) I'll add it back, but I will have to find an other UI trick to make it interact as a "readonly" property

amrsoll:
For future reference: an old StEx post describes that properties that only have a `get` function can be seen as greyed out in the UI.

https://blender.stackexchange.com/a/13702

So either I

- destruct+rebuild the property each time it swaps to/from readonly state, but that breaks drivers and also other code that might have outdated refs to the object
- edit the Property using the function described earlier, but I still don't get how to find the "path" of the Property to edit.

hlorus:
> For future reference: an old StEx post describes that properties that only have a `get` function can be seen as greyed out in the UI.
> 
> https://blender.stackexchange.com/a/13702
> 
> So either I
> 
> - destruct+rebuild the property each time it swaps to/from readonly state, but that breaks drivers and also other code that might have outdated refs to the object
> - edit the Property using the function described earlier, but I still don't get how to find the "path" of the Property to edit.

Shouldn't the layout.enabled flag be the solution here?
https://docs.blender.org/api/current/bpy.types.UILayout.html#bpy.types.UILayout.enabled

Maybe the StEx post is just too old so it wasn't suggested there.

amrsoll:
> Shouldn't the layout.enabled flag be the solution here?

The problem is that is disable the whole UILayout in question, column, row... not individual props inside that element.

If the readonly attribute is defined in the Property, it is way more portable and better separation of concerns.

hlorus:
> The problem is that is disable the whole UILayout in question, column, row... not individual props inside that element.
> 

A prop could still be wrapped in it's own sub row..

> If the readonly attribute is defined in the Property, it is way more portable and better separation of concerns.

True, also it seems to break the drivers workflow...

I would say a conditional in the setter is the best, least hacky approach.

amrsoll:
> A prop could still be wrapped in it's own sub row

I reused the row segments that were available in the menu

> conditional in the setter is the best

That cannot be used because We are also updating the value with the entity size. However, because it gets automatically updated, it is not the end of the world if the user is able to input a value because it will get updated if it isn't solver-compatible.

amrsoll:
And I just received the steam deck, so productivity fell through the floor :joy:

amrsoll:
I am looking forward to read your suggestions for improving the code structure :)

amrsoll:
I'm just sad that you cannot create a driver if the UI element is disabled, but it makes sense not to write to that value through the driver. I guess that is potential bug when not having a conditional limiter in the setter function.

hlorus:
> That cannot be used because We are also updating the value with the entity size. However, because it gets automatically updated, it is not the end of the world if the user is able to input a value because it will get updated if it isn't solver-compatible.

We could actually set the property value without triggering the getter/setter/update callbacks, at least from within the constraints classes.



> I'm just sad that you cannot create a driver if the UI element is disabled, but it makes sense not to write to that value through the driver. I guess that is potential bug when not having a conditional limiter in the setter function.

IMO it makes sense that it's not possible to add drivers, what's missing is the "copy as new driver" and "driver datapath" options i suppose...

amrsoll:
I didn't want extra functions to update the private var, but I suppose I can add the extra argument `force=False` to allow force updating using the same function.

amrsoll:
Also please let me know about cosmetic changes, I re-read that you were thinking of having the value in brackets, maybe we want to dim the color of the font etc...

ui.py
```diff
@@ -272,28 +306,32 @@ def draw_constraint_listitem(context, layout, constraint):
...
+    center_sub.alignment = "LEFT"
 
     # Dimensional Constraint Values
     for constraint_prop in constraint.props:
-        sub.prop(constraint, constraint_prop, text="")
+        center_sub.prop(constraint, constraint_prop, text="")
+
+    # Disable interaction with element if it is "readonly"
+    center_sub.enabled = (
+        isinstance(constraint, DimensionalConstraint) and constraint.is_reference
```
hlorus, line 326: This seems to be inverted, disables constraints that are not only reference and allows editing reference constraints.

hlorus (commented):
- It seems like the color changes are breaking the constraint highlight when an operator button is hovered (when its tooltip appears) in the interface. Maybe have a look at the HighlightElement mixin class.
- Did you try adding a driver so one constraint has the value of a reference constraint? It's quite a bummer that you cannot use "copy as new driver", do you now if there's another simple way for users to add such a driver?
- Maybe @exegetor could also have a quick look at the checnges regarding the init_props / assign_init_props methods...?

model/base_constraint.py
```diff
@@ -173,3 +174,69 @@ def index(self):
...
+            # Could not find a way to have the property "readonly",
+            # so we disable user input instead
+            col.prop(self, "value")
+            col.enabled = not self.is_reference
+        if hasattr(self, "setting"):
+            row = sub.row()
+            row.prop(self, "setting")
+        return sub
+
+    def is_active(self, *args, **kwargs):
```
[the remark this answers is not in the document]
amrsoll: Correct, I thought I had deleted that thing :+1:

ui.py
amrsoll: True, fixed :)

amrsoll:
> It seems like the color changes are breaking the constraint highlight when an operator button is hovered (when its tooltip appears) in the interface. Maybe have a look at the HighlightElement mixin class.

How so? To me the highlight colors display correctly. What UI item are you hovering when it should highlight? I tried to compare with main without any different behaviour w/regards to highlighting colors.

> Did you try adding a driver so one constraint has the value of a reference constraint?

Had to look up how to do that with MakerTales video :eyes: but yes! it works only if

1. You create the driver from constraint A
2. Make the constraint A reference / readonly
3. Paste driver into Dimensional constraint B

amrsoll:
Example test file for the feature :)

[untitled.blend.zip](https://github.com/hlorus/CAD_Sketcher/files/9551228/untitled.blend.zip)

amrsoll:
2 things about re-enabling the UI elements:

- Even when enabled in the constraint context menu (right click on constraint), the dimension cannot be copied as a driver
- If re-enabling all constraints in the Constraint UILayout list, then
  - there is no way to differentiate between the reference and normal constraints
  - You can create the driver at anytime from a reference item, no need to switch reference mode :)

amrsoll:
Side note: create a StEx question for the extra spicy solution: deleting/recreating the setter function

https://blender.stackexchange.com/questions/274411/override-edit-the-get-set-functions-of-a-property

model/base_constraint.py
```diff
@@ -173,3 +174,74 @@ def index(self):
...
+        if not self.get("value"):
+            self.assign_init_props()
+        return self["value"]
+
+    def assign_init_props(self, context: Context = None):
+        # self.value, self.setting = self.init_props()
+        _value, _ = self.init_props()
+        self._set_value_force(_value)
+
+    def refresh_constraint(self, context: Context = None):
```
hlorus, line 213: Might be great to name this so it reflects that it's only used with reference dimensions...

hlorus:
> > It seems like the color changes are breaking the constraint highlight when an operator button is hovered (when its tooltip appears) in the interface. Maybe have a look at the HighlightElement mixin class.
> 
> How so? To me the highlight colors display correctly. What UI item are you hovering when it should highlight? I tried to compare with main without any different behaviour w/regards to highlighting colors.
>

Oops, you are right! The highlight color was just too subtle, sorry about that.
 

> I'll make the change with the `force=False` variable and I'll let you see if that is preferable to making the UI element grayed out.

IMO we have to go this route, it's too annoying if you cannot create the driver when a constraint is already set to reference. There needs to be some kind of distinction between reference and normal constraints in the sidebar indeed, not sure how exactly tho, maybe just a separate block with a title "Reference:" ?

hlorus:
> 2 things about re-enabling the UI elements:
> 
> * Even when enabled in the constraint context menu (right click on constraint), the dimension cannot be copied as a driver
> * If re-enabling all constraints in the Constraint UILayout list, then

Yeah thats a unfortunate limitation, blender doesn't allow to add drivers from popup menus, that's why we need to display the properties in the sidebar....

> On a different note, **pasting a driver onto a reference constraint doesn't change the size of the reference measurement**. After 2 sec of reflection, that is logical, but will probably confuse users.

Yeah ideally the disabled layout would allow to copy as new driver but disallow adding a driver... but i think we found a ok workaround for the moment.

solver.py
```diff
@@ -122,7 +122,7 @@ def _get_msg_entities():
             # Store a index-constraint mapping
             from collections.abc import Iterable
 
-            indices = c.create_slvs_data(self.solvesys, group=group)
+            indices = c.py_data(self.solvesys, group=group)
```
hlorus, line 125: That's quite the fundamental change here, is it really needed for this PR? I would rather have this in a separate PR and apply it also to entities.

But i'm not actually sure this makes sense, The slvs solve system cannot be created and updated at a later point. So a new one is created everytime the solver runs. If we don't explicitly recreate slvs data py_data might return handles from the last solver run.

hlorus:
Regarding cosmetics, adding brackets and dimming the text sounds reasonable but i don't have a strong opinion here.

solver.py
amrsoll: I am going to think of a different way to do it so it doesn't have that many changes to the higher level abstractions

amrsoll:
> blender doesn't allow to add drivers from popup menus

Then I will re-disable the value selection in the popup menu

> ideally the disabled layout would allow to copy as new driver but disallow adding a driver...

I haven't tested this, but I believe that a custom Property that only has a `get()` function and not `set()` would have that behaviour.

> Regarding cosmetics, adding brackets and dimming the text sounds reasonable but i don't have a strong opinion here.

I think we should run a poll in the discord to gauge what people think

solver.py
amrsoll: @hlorus 
Either:
- I rename the `create_slvs_data()` in all the constraints and entities that use it.
- I write the "empty return" return step in all the `create_slvs_data()` for the dimensional constraints.

The reason I refactored `create_slvs_data()` to `py_data()` in the higher level abstraction is because the `.py_data` attribute is used by the entities to pass on the solver handles.

I can create a separate PR now that refactors all the entities and constraints to 
- have a `py_data()` function that serves as interface and returns the slvs handles.
  > That function should only be implemented by the abstract/mother classes like `GenericX` 
- have a private function `create_slvs_data()` that generates the data.
  > This function is implemented by all the daughter classes to create handles specific to themselves

I would also suggest a name change for those functions, for example `solver_handles()` and `generate_solver_handles()`

hlorus:
The functionality of init_props is changed here, if i get it right calling assign_init_props would now have the same effects as init_props had before. Yet there are places (operators/base_constraint.py line 69) where init_props is still called. While i haven't tested it i'm afraid this breaks stuff, was it intentional?

amrsoll:
It was intentional, but I feel like I need to open a discussion topic so we don't pollute the PR too much.

hlorus:
> It was intentional, but I feel like I need to open a discussion topic so we don't pollute the PR too much.

Okay then i'd say we can merge this, i haven't found anything braking. Maybe just keep an eye open for reports.

vespakoen:
@amrsoll Looks like this PR broke the Angle constraint ;)

vespakoen:
More specifically, this commit: https://github.com/hlorus/CAD_Sketcher/commit/ee09f49ec6e2653109ae58a29df2ede2997651e0

PR #284, opened 2022-09-01
adds debugpy python lib dependency

therealjacobbrauer:
this edit adds in the necessary to install dependency of ```debugpy``` to the dependencies section of the readme.


hlorus:
Thanks for the PR. As this is something that only affects developers i won't put this inside the readme. Better have it here: https://github.com/hlorus/CAD_Sketcher/blob/main/docs/advanced.md
Under contribute -> development

therealjacobbrauer:
Thanks for the suggestion @hlorus about the preferred location of the developer dependency. I added it to the file that you suggested.

PR #309, opened 2022-10-31
Sinplified func(self)

Nex-Pro:
simplified code using ternary ( not tested )

### DIFF
--- a/model/utilities.py
+++ b/model/utilities.py
@@ -20,7 +20,7 @@ def slvs_entity_pointer(cls, name, **kwargs):
     @property
     def func(self):
         index = getattr(self, index_prop)
-        return none if index == -1 else bpy.context.scene.sketcher.entities.get(index)
+        return None if index == -1 else bpy.context.scene.sketcher.entities.get(index)
     setattr(cls, name, func)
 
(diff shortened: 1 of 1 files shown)


hlorus:
Thanks for the patch, i agree the snippet of code could be imporved. However putting everything on one line doesn't make it more readable IMO. I would suggest to simply remove the else as it's not needed.

model/utilities.py
```diff
@@ -20,11 +20,7 @@ def slvs_entity_pointer(cls, name, **kwargs):
     @property
     def func(self):
         index = getattr(self, index_prop)
-        if index == -1:
-            return None
-        else:
```
hlorus, line 25: Simply removing this line and unindenting the else block improves readability alot.

model/utilities.py
```diff
@@ -20,11 +20,7 @@ def slvs_entity_pointer(cls, name, **kwargs):
...
-            return bpy.context.scene.sketcher.entities.get(index)
-
+        return none if index == -1 else bpy.context.scene.sketcher.entities.get(index)
```
amrsoll, line 23: *None :)

PR #320, opened 2023-01-13
Add new solvespace solver

vespakoen:
Hi there!

I am Koen, and I am a maintainer of SolveSpace, thanks for this addon, it looks really cool!

My goal is to merge Yuan Chang's PR (https://github.com/solvespace/solvespace/pull/493) that adds the python bindings to SolveSpace into the main branch, and make sure it works and keeps working in the future.

I also renamed the package from `python_solvespace` to `solvespace`, and worked on the CI a bit.
That code is over here in my fork: https://github.com/vespakoen/solvespace/tree/python

From this fork I published a test to test.pypi.org over here: https://test.pypi.org/project/solvespace/#files

I am looking for feedback, and would like the package to be tested on as many platforms as possible.

This PR is based on [#268](https://github.com/hlorus/CAD_Sketcher/pull/268) but based on the latest main branch, it fixes an issue in `solver.py`, where the `self.solvesys.set_group(group)` was set in the wrong place.

This is not complete yet, but it seems to work a little at least, and I hope that it will move things along ;)

To try it out:
```shell
mkdir -p blender/scripts/addons
cd blender/scripts/addons
git clone -b solvespace-solver https://github.com/vespakoen/CAD_Sketcher.git
/path/to/blenders/bin/python -m pip install -i https://test.pypi.org/simple/ solvespace==3.1.0.dev43 --force-reinstall
# - add blender/scripts path to Blender Preferences -> File Paths -> Scripts
# - activate the extension
# - restart blender
```

For the curious, it seems to fix the angle bug?

Screenshot 2023-01-13 at 02 35 09

Breaks a bunch of other things though because it is incomplete ;)


vespakoen:
Oops, looks like I was actually using an old fork, will rebase...

vespakoen:
Fun fact, the angle issue is back after rebasing, so it might make sense to add a failing test case and git bisect to see where it got introduced =)

amrsoll:
Thank you for the PR, very cool :D Yeah I managed to introduce a separate bug with the angles unfortunately. Usually it makes the angles switch between the direct or the complementary angle. 

The original problem was that the solver just didn't want to produce a solution, which was why I thought it was best to have an newer or a different solver :) of course, it is a lot of work to update the API endpoints.

I would suggesting converting to a "draft" PR if it is not functional yet. ;)

hlorus (commented):
Hey Koen, 
Great work, it's nice to see effort on an official python binding! Definitely looking forward to having this PR merged.

This will probably have to wait until the package is available on the official PyPi, in the meantime we could have these changes in a branch to work on the remaining issues and to get people to test it.

vespakoen:
Thanks for the suggestions, and yes, I will merge the python-solvespace bindings once I am confident that they are working well, so basically I want to make sure that the compiled libraries are working on all platforms, and that all the functions are working as they should.

Thanks to @amrsoll's work, I now have a nice way to test the bindings =)

vespakoen:
Note: to use the latest fixed, you have to install the pre-release of the solvespace python package:

```
/path/to/blender/python -m pip install -i https://test.pypi.org/simple/ solvespace==3.1.0.dev43 --force-reinstall
```

vespakoen:
For the horizontal and vertical points constraints, you need a newer version of the solver:

```shell
/path/to/blender/python -m pip install -i https://test.pypi.org/simple/ solvespace==3.1.0.dev44 --force-reinstall
```

vespakoen:
It all seems to work pretty well now, and all the tests I added are passing =) would love some feedback from others that have used the main CAD Sketcher a lot and can maybe find differences more easily.

hlorus:
Awesome, i'd suggest to ask people on discord for testing. You should have been added as a collaborator on this repository so feel free to create a branch to have people test it.
I'll also have another look at it in the coming days. :)

hlorus:
There are a few things i've noticed that don't work so far:
- Tangent between two curves
- Equal between a line and an arc
- Vertical/Horizontal aligned distance

Let me know if there's something i can help with.

phkahler:
@vespakoen  I haven't been following the solver branch or CAD sketcher very closely. Two major changes have happened in Solvespace and I wonder if they affect this much.

1) Last year we (mostly Ryan via Evil Spirits work) integrated Eigen, which introduces an external dependency.
2) Just recently we added the multi-constraint patch, fixed, and expanded it.

I'm not sure where the external solver API sits, so maybe that second item has no effect, or maybe it's really good news for CAD sketcher. You tell me ;-)

I also added faces parallel and perpendicular, but I don't think CAD sketcher is using 3D constraints is it?

Thanks!

vespakoen:
@hlorus Thanks for the tests, I have actually also found those myself, and a couple of other problems as well.

Basically, the `python-solvespace` and `slvs_py` bindings wrap the C library that SolveSpace exposes, this C library is actually missing some good parts of SolveSpace, for example, and most notably: https://github.com/solvespace/solvespace/blob/master/src/constraint.cpp#L134-L195

Also, the C library doesn't work in the most efficient way, basically CAD Sketcher is loading all the drawings into the SolverSystem on every solve, and I guess, when dragging something, this might slow things down at some point.

The `python-solvespace` library adds it's own "Entity" class, that wraps a SolveSpace EntityBase, but it is not "the actual entity", this causes users to keep track of things themselves, which can actually be avoided when the SolveSpace library would be C++ based.

So the last 2 days I went down into the rabbit hole of making pybind11 bindings for SolveSpace, and for that I first had to "rip out" the core of SolveSpace, many things depend on each other, and some SolveSpace UI specific stuff is also living in the "core" parts.

The most notable offender here is the "Group" object, it contains a lot of code for UI / Mesh related things, but the Solver can work with a group that only has a couple of properties and a single method.

My goal now is to "untangle" the core from SolveSpace, make that available as a C++ library and as pybind11 python bindings, this will be a net positive for other applications using the Solver (like CAD Sketcher) because they don't have to repeat the work that is already done in SolveSpace, and also for SolveSpace, because the "core" is seperated from the UI stuff.

Basically, I have added a bunch of methods to the "Sketch" object (basically the ones that are on python-solvespace's SolverSystem object (added it to a gist for now: https://gist.github.com/vespakoen/2abe264d0f12736659cf9eb19f2daed5) 

And some to the `EntityBase` (also "stolen" from the python-solvespace project):

```c++
    inline bool IsPoint2D() {
        return type == Type::POINT_IN_2D;
    }

    inline bool IsPoint3D() {
        return type == Type::POINT_IN_3D;
    }

    inline bool IsNormal2D() {
        return type == Type::NORMAL_IN_2D;
    }
    ... etc
```

This is all very experimental at the moment, but I am hoping this can work out the way I imagine it, if not, I will make sure to add the missing stuff to the python-solvespace library.

@phkahler The Eigen dependency is not a big problem, it's header only and can easily be included, the multi-constraint stuff is not exposed in the library, this is something that CAD Sketcher has to implement, and in some ways already did I think.

gerion0:
Just to save others the research work: @vespakoen [merged](https://github.com/solvespace/solvespace/commit/a208201c32d52fac366da74e21dfb3ae4d8eff65) the Python bindings in solvespace upstream. If I see it right, this is the first big step of the PR. Thank you!

hlorus:
Thanks for this. The addon has since gone through a major rework with native Blender Curves as the data source, and this predates that change, so closing as no longer applicable. Feel free to open a fresh PR if you would like to revisit it.

PR #322, opened 2023-01-17
Fix set_value for complementary angle constraint.

### LINKED ISSUE #310: [BUG]  Supplementry Angle Unit Input Doing strange Things
### Contact Details

_No response_

### Description

When you try and control/input an angel to a supplementary angle things go a little weird. @hlorus hase taken a small look and said that it looked slightly "random" might be a good call for @amrsoll to take a look at it as you have the most experience working on this area of the code,

Capture (1)

### Addon Version

latest

### Blender Version

3.3

Windows

amrsoll:
/!\ This might break some save files, as angles measured as "complementary" will be displayed as 180 - their.value 

This is a consequence of one of my previous problematic PRs, where the value of the constraints always refer to the same physical attribute. (e.g. the direct angle between 2 lines)

Although it introduces more functions, I still believe the idea is sound though, as we can guarantee that my_angle_property["value"] always refers to the direct angle.

Perhaps the functions `to/from_displayed_values()` do not fit in the class `DimensionalConstraint` because it should not be concerned with the UI. Please let me know if there is a better place to have it.

Hopefully fixes #310 please test on your side too :see_no_evil: 

N.B. Fixes a corner case where an initial direct angle of value 0 would not display properly.

### DIFF
--- a/model/angle.py
+++ b/model/angle.py
@@ -27,11 +27,6 @@ class SlvsAngle(DimensionalConstraint, PropertyGroup):
     The constraint's setting can be used to to constrain the supplementary angle.
     """
 
-    def _set_value_force(self, value):
-        DimensionalConstraint._set_value_force(
-            self, HALF_TURN - value if self.setting else value
-        )
-
     def assign_init_props(self, context: Context = None):
         # Updating self.setting will create recursion loop
         _value, _ = self.init_props()
@@ -69,6 +64,9 @@ class SlvsAngle(DimensionalConstraint, PropertyGroup):
         return WpReq.NOT_FREE
 
     def to_displayed_value(self, value):
+        return HALF_TURN - value if self.setting else value
+
+    def from_displayed_value(self, value):
         return HALF_TURN - value if self.setting else value
 
     def create_slvs_data(self, solvesys, group=Solver.group_fixed):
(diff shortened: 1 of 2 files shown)


hlorus:
Thanks for looking into this! It seems to indeed fix the bug in #310. Regarding the breaking of files: there's versioning support in the addon already, so just bump the addon version in the bl_info and add something like the following to versioning.py:
        if version < (0, 26, 0):
            for c in context.scene.sketcher.constraints.angle:
                if not c.setting:
                    continue
                msg += "\n Recalc supplementary angle of: {}".format(repr(c))
                c.assign_init_props(context)

That being said there's still some jumping around when adding a new angle constraint, not sure if that was the case before as well but might be related...

amrsoll:
Unfortunately I am not a 100 percent sure if I can write a migration step that can fix the angles in the save file given that it was flaky behaviour beforehand :/ 

The jumpiness when creating angle constraints is probably related. Unfortunately I don't know if I will have time in the near future to look more into it... Perhaps reverting some of my changes for handling angles is the easiest.

Note: I find that breakpoints on the set_ functions are easier to debug with given that the get_ functions get called constantly :)

hlorus:
> Unfortunately I am not a 100 percent sure if I can write a migration step that can fix the angles in the save file given that it was flaky behaviour beforehand :/ 
> 

assign_init_props would simply recalculate the constraint value which currently might  cause jumping but this could still be better than doing nothing..

PR #323, opened 2023-01-23
Solvespace solver using pybind11 module

vespakoen:
This is based on #320, but changes some things to make it work with my (very) experimental pybind11 bindings of the SolveSpace solver.

That code lives over here: https://github.com/vespakoen/solvespace/tree/experiments/src/core

I basically stripped out the solver code from SolveSpace and organised it into smaller files.
I added constraint / entity helper methods in the Sketch class (https://github.com/vespakoen/solvespace/blob/experiments/src/core/sketch.h#L25-L652)
A `SolverSystem` class (https://github.com/vespakoen/solvespace/blob/experiments/src/core/solversystem.cpp)
And a `pybind11` bindings file (https://github.com/vespakoen/solvespace/blob/experiments/src/core/main.cpp)
Which exposes way to many functions at the moment, but whatever...

The good parts:
- Compared to `python-solvespace`, this adds the helper methods in a C++ class so it can be re-used, possibly by other bindings in the future.
- The `other` and `other2` properties for the constraints are handled inside of this library (see https://github.com/vespakoen/solvespace/blob/experiments/src/core/sketch.h#L557-L628) this makes tangent constraints work correctly.
- Direct access to SolveSpace internals, allowing you to change properties and call methods on the "actual objects" that are on the sketch, allowing for future speed improvements in CAD Sketcher because it is possible to solve without copying stuff around so much, and possibly make implementing some functionality easier.

The bad parts:
- I ripped out the core from SolveSpace, making SolveSpace use this "new" core is going to be a substantial amount of work, and because the change / PR will be a big one, the chances of it getting accepted are more slim, perhaps this has to live on in a fork until it's proven to be stable / useful.
- Memory lifecycle management hasn't been tested / taken care of, this is an experiment a whipped up over a couple of days,  getting something working was the first and most important step, although I think this shouldn't be too big of an issue.
- This code probably only works with a single sketch, because I made some bigger changes to the `solver.py` in CAD Sketcher, and it's not complete yet.

In general, I would love this branch "to succeed", because I believe it has the most potential for CAD Sketcher, and I believe that having the Solver code more separated from the rest of SolveSpace is also a net positive, it does however come at a cost (more work).

Interested to hear your thoughts ;)

This branch can be tested out with the following python package:

```
pip install --force-reinstall -i https://test.pypi.org/simple/ solvespace==3.1.0.dev53
```


hlorus:
Is this version of the python binding also available on the pip testserver?

It definitely sounds great from the addon perspective as the cons you've listed affect the solvespace side. I'd say it would be good to know from solvespace devs if this has a chance to get merged. A fork might be alot of extra work and has the risk of being discontinued at some point which would be annoying.

vespakoen:
Yes, I pushed it is on PIP with version dev53 (pip install --force-reinstall -i https://test.pypi.org/simple/ solvespace==3.1.0.dev53)
I experimented some more with this branch, and tried to integrate it back into SolveSpace, but it is really, really hard, I am afraid I will have to give up on this or find help with it, this is probably better suited for the far future, or a fork.

I will now try if I can bring the "good parts" into the python-solvespace branch.

Aveline67:
I did everything needed but I get "Module isn't registered" in Blender 3.4. How can solve this ? I really want to test this branch

vespakoen:
You probably didn't install the correct version of the solvespace package?
I have been publishing different kinds of bindings under the solvespace name.

The Pybind11 bindings got abandoned, but the new bindings are Cython based, the API and functionality is pretty much the same but it is a bit faster and needs less changes made to SolveSpace. I am hoping to release that soon and will also make a new PR to this project that uses the newest bindings.

The new bindings are also "stateful" which basically means that the "sketch" objects do not get freed/cleaned up. and in theory will allow one to re-use it / add / replace / remove items on and from the sketch. There is however no API yet to replace or remove entities and constraints at the moment.

However, when that gets added it will be possible to make a more performant CAD Sketcher because you don't have to re-create / send all the sketches

PR #328, opened 2023-01-25
Blender 3.4.1 now says "PIP" instead of PyPi

petemyron:
Neither are wrong, but it could be confusing to non-Python nerds, so it's probably best to state what's actually shown there.


hlorus:
Thanks!

PR #333, opened 2023-02-15
Fix installing addon directly using GitHub zip

Cheaterman:
Use relative imports all the way down

### DIFF
--- a/ui/__init__.py
+++ b/ui/__init__.py
@@ -1,7 +1,10 @@
 import bpy
 from bpy.types import Context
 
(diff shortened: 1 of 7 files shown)


hlorus:
So blender indeed has a problem with absolut imports?

Cheaterman:
Sadly not so much Blender, but our installation instructions :-) we recommend people drag'n'drop the zip that comes directly from GitHub, sadly this zip calls the top level folder `Cad_Sketcher-branchname` so that will indeed break absolute imports as-is!

Issue #218, opened 2022-07-07
Code Cleanup


hlorus:
As the addon has grown quite a bit and new contributors are jumping onboard there's some cleanup overdue. This task is used to keep track of code cleanup efforts.

Feel free to suggest additional topics/areas in the comments.

## Structure
- [x] #219
- [x] Improve registration flow #223
- [x] Slip out utilities and constants https://github.com/hlorus/CAD_Sketcher/commit/b73d41b1ce5749abb988533e4bb6f215e468353e#diff-8656a273c89d5812379185ade358273fa3d93e2cb8df832c66cec928eef9809a
- [x] Split out blender classes into into individual files (PropertyGroups, Operators) #229
- [x] Split out UI definitions into individual files (ui.py, workspacetools.py)
- [x] Move all keymap definitons to keymaps.py
- [x] Split out gizmos.py into individual files

## Style
- [x] #248 

(docstrings and type hints tracked in #335 and #171 respectively)

bonjorno7:
It helps a lot with maintaining readability if everyone formats their code the same way.
I use YAPF for formatting, which can be setup in most popular text editors.
Here are my VS Code settings for reference:
```json
{
    "editor.rulers": [
        80,
        120
    ],
    "editor.formatOnPaste": true,
    "editor.formatOnSave": true,
    "python.formatting.provider": "yapf",
    "python.formatting.yapfArgs": [
        "--style",
        "{based_on_style: google, column_limit: 120}"
    ],
    "python.sortImports.args": [
        "-l",
        "120"
    ],
    "[python]": {
        "files.insertFinalNewline": true,
        "files.trimFinalNewlines": true
    },
    "python.linting.enabled": true,
    "python.linting.pylintEnabled": false,
    "files.exclude": {
        "**/*.pyc": true
    }
}
```

hlorus:
> It helps a lot with maintaining readability if everyone formats their code the same way.
> I use YAPF for formatting, which can be setup in most popular text editors.
> Here are my VS Code settings for reference:
> ```json
> {
>     "editor.rulers": [
>         80,
>         120
>     ],
>     "editor.formatOnPaste": true,
>     "editor.formatOnSave": true,
>     "python.formatting.provider": "yapf",
>     "python.formatting.yapfArgs": [
>         "--style",
>         "{based_on_style: google, column_limit: 120}"
>     ],
>     "python.sortImports.args": [
>         "-l",
>         "120"
>     ],
[...]

PR #334, opened 2023-02-15
Split workspacetools.py, move some keymaps to keymaps.py

Cheaterman:
This should address the splitting of workspacetools.py in #218.


workspacetools/select.py
```diff
@@ -0,0 +1,117 @@
...
+
+class VIEW3D_T_slvs_select(WorkSpaceTool):
+    bl_space_type = "VIEW_3D"
+    bl_context_mode = "OBJECT"
+    bl_idname = WorkSpaceTools.Select
+    bl_label = "Solvespace Select"
+    bl_description = "Select Solvespace Entities"
+    bl_icon = "ops.generic.select"
+    bl_widget = GizmoGroups.Preselection
+    bl_keymap = (
```
hlorus, line 15: Maybe we could move these keymap items to keymaps.py as well..

workspacetools/add_arc2d.py
```diff
@@ -0,0 +1,22 @@
...
+
+class VIEW3D_T_slvs_add_arc2d(GenericStateTool, WorkSpaceTool):
+    bl_space_type = "VIEW_3D"
+    bl_context_mode = "OBJECT"
+    bl_idname = WorkSpaceTools.AddArc2D
+    bl_label = "Add 2D Arc"
+    bl_operator = Operators.AddArc2D
+    bl_icon = "ops.gpencil.primitive_arc"
+    bl_widget = GizmoGroups.Preselection
+    bl_keymap = (
```
hlorus, line 17: These lines are repeated for every tool, maybe we could move it to keymaps.py and just import one keymap to avoid potential inconsistencies between tools.

hlorus (approved):
Generally looks good to me, see inline comments.

PR #337, opened 2023-02-17
Add icons to "Add constraints" panel, refactor icon_manager accordingly

### LINKED ISSUE #160: Constraint Icons in N-Panel
### Discussed in https://github.com/hlorus/CAD_Sketcher/discussions/122

<div type='discussions-op-text'>

<sup>Originally posted by **coleorloff** May 10, 2022</sup>
**Motivation / Existing Problem**
It's a little difficult to "read" sketches without familiarity with the constraint icons. (It's also a Blender UI convention to have icons paired with menu items.)

**Description**
Add constraint icons next to their name in the CAD Sketcher N-panel. T

his would provide a positive feedback loop in terms of visually identifying symbols within the system and "reading" your sketches, both for beginners and more advanced users.

**Mock-Up**
Note: highlighted area shows a selection of constraints paired with their icons. The goal would be to have all constraints in the menu paired with their icon. This is only for illustrative purposes. 

CADSketcher_N-panel_proposal

</div>

Implantation :
- How custom icons can be added to the interface is documented [here](https://docs.blender.org/api/current/bpy.utils.previews.html#bpy.utils.previews.ImagePreviewCollection)
- It would be great to use the same icons in the 3d viewport to avoid confusion. Therefore it would be nice if icon_manager.py could read the icon data that is used for the interface.

Cheaterman:
Closes #160


icon_manager.py
```diff
@@ -72,10 +98,50 @@ def load():
...
+
+    constraint_icons = bpy.utils.previews.new()
+
+    for operator, type in _operator_types.items():
+        icon_path = get_folder_path() / f"{type}.png"
+
+        if not icon_path.exists():
+            continue
+
+        constraint_icons.load(operator, str(icon_path), 'IMAGE')
```
hlorus, line 121: Actually kinda strange to depend on rhe operators, couldn't we just convert all files in rhe folder and name the .dat's according to the filename?

Cheaterman: It's deterministic - we got bitten enough by magic, at this point I'd rather be explicit and correct, haha. Plus, loading PNG files is actually fairly heavy, so if there were eg extras in the folder that we wouldn't want to load, this would be a bad idea anyway.

Cheaterman: Turns out it _was_ already expensive enough to load the PNGs, I had to load the preview icons instead since they're more than big enough ; used to be 120ms per image, now is 9 (not a typo)

ui/__init__.py
```diff
@@ -48,6 +49,8 @@ def draw_add_sketch_in_add_menu(self, context: Context):
 
 
 def register():
+    icon_manager.load_preview_icons()
```
hlorus, line 52: Shouldn't this be called from the icon_manager.load() function? Or at least from the same place..

Cheaterman: Absolutely, but there's a combination of reasons I left it here - icon_manager doesn't have an unload() function so I couldn't unload_preview_icons() there, and it felt asymmetric and inconsistent to have only unload_preview_icons() in `ui/__init__.py` without load (inevitably someone in the future will think it was an accident and will re-add it), so I figured the best course of action was to make load_preview_icons() idempotent and call it in both places. If you think one of the other alternatives above is better, or that I should add an unload() to icon_manager, let me know! I will merge this for now but we can always add more commits to main branch.

hlorus: An unload function in icon_manager.py that is called from the root init file's unregister function seems reasonable to me 👍🏼

Cheaterman: Done in  8f67f30 :+1:

PR #338, opened 2023-02-18
Cleanup: Remove operator dependency in icon_manager

hlorus:


Cheaterman (approved):
I don't think a glob() is the way to go until we solve the performance issues (and allow uploading more than 32² textures), but up to you.

PR #340, opened 2023-02-22
Extend addon installation guide for mac users

hlorus:


Cheaterman:
A user reported this today too, I think we might want to merge it :-)

PR #341, opened 2023-02-26
Fix resource folder typo

coopbri:
- Renamed all instances of 'ressources' to 'resources'


hlorus:
LGTM, thanks!
Maybe @Cheaterman wants to have a quick look at this since he worked on this area lately?

Cheaterman:
Looks good to me too, thanks :-)

ddavace:
If you get really unlucky and have a version before this change installed, and then try and remove and install a version after this change.  Then you end up like I did.  This is just a comment for the next person:

The solution is easy, just restart blender AFTER you remove the add-in and BEFORE you install the new one.
(certainly doesn't seem serious enough to warrant any further action though.)

PR #351, opened 2023-03-01
Ui: Add context menu for selected entities

hlorus:
- Menu is called on right-click when nothing is hovered
- Common properties are shown: visible, construction, fixed
- Icon indicates if property on all entities is True/False ore mixed
- If values are mixed the operator will set the property on all entities to a default


hlorus:
This now also displays the use_construction property (added by @SiboVG) when nothing is selected. 

Currently it might not be clear to the user what the button/icon means as it could either display the state of the property on the selected entities or the construction mode toggle. Therefore we could think about tweaking the icon or maybe using the blue button highlight...

@TheMakerTales maybe you wanna give this a go and see how ot feels so far?

PR #379, opened 2023-04-11
Optimise UI and better resize behaviour

alexfacciorusso:
The names and values of the constraints and entities had a fixed ratio, which meant when the panel was larger, both names and values of the rows were not expanding, effectively wasting space that could have been used to show information.

If it this had already been done and there's any reason why wouldn't we have these changes, let me know!

| Before | After |
|--------|--------|
| https://user-images.githubusercontent.com/4479527/231173541-51af97d9-b19c-4abb-a9f2-ef8b332f3c4d.mp4 |  https://user-images.githubusercontent.com/4479527/231172428-d22244bd-cb00-491d-bec0-45e2dea2148d.mp4 |

### DIFF
--- a/model/diameter.py
+++ b/model/diameter.py
@@ -18,7 +18,6 @@ from ..utilities.solver import update_system_cb
 
 logger = logging.getLogger(__name__)
 
-
 class SlvsDiameter(DimensionalConstraint, PropertyGroup):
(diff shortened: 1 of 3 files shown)


hlorus:
Looks goot to me, thanks for the PR!

Issue #50, opened 2022-04-27
Overlapping entity selection


hlorus:
Somehow allow selecting overlapping entities

TheMakerTales:
This issue has been given a $300 bounty!
Find out more about bounties here: [https://www.cadsketcher.com/cad-sketcher-bounty-board](https://www.cadsketcher.com/cad-sketcher-bounty-board)

hlorus:
Resolved by the selection rewrite: picking is now CPU analytic (screen-space) instead of the GPU id-buffer, which was the cause here (the id-buffer could only hold the topmost entity per pixel). Box-select now tests every entity independently and grabs overlapping ones (regression test: test_box_selects_overlapping). Please retest on the latest prerelease build. Reopen if it persists.

PR #387, opened 2023-05-25
Early WIP for entity multi selection

Cheaterman:
Need more eyes and hands on this, both for review and for testing :-) - it's still VERY buggy and somewhat incomplete ; documentation also needs to be adjusted, and some docstrings in the (admittedly very complex) code are sorely needed too.

### DIFF
--- a/draw_handler.py
+++ b/draw_handler.py
@@ -6,6 +6,7 @@ from bpy.types import Context, Operator
 from bpy.utils import register_class, unregister_class
 
 from . import global_data
+from .utilities.index import update_selection_map
 from .utilities.preferences import use_experimental
 from .declarations import Operators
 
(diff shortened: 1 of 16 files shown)


Cheaterman:
This is meant to address #50 BTW :-)

hlorus:
This looks great as far as i've looked.

What exactly is very buggy, the core selection system itself or the adoption in the rest of the addon's operators?

Also how ar the entities in the hover list sorted? IMO defaulting to always showing a list of entities when selecting is bad as this would be the case for every line endpoint. Therefore it would be great if we could be sure that the first entity in the list is a reasonable choice and just pick that by default. Then use alt+lmb for the list.

As far as th "reasonable decision" goes, currently points are just always drawn before lines/circles etc which seems to work ok in most cases.

hlorus:
This doesn't seem to work on my system. Usually there's just one entity highlighted or suddenly all entities are highlighted. Also i'm getting the following traceback when creating a simple rectangle and hovering an edgepoint:

KeyError: 14
  File "/home/yoga/Dropbox/blend/addons/CAD_Sketcher/gizmos/preselection.py", line 67, in test_select
    indices = rgba_to_indices(r, g, b, alpha)
  File "/home/yoga/Dropbox/blend/addons/CAD_Sketcher/utilities/index.py", line 79, in rgba_to_indices
    indices.append(_selection_to_index[index])

hlorus:
Thanks for this. Multi-selection is now handled (box-select plus overlapping-selection cycling in #614), so closing this WIP as no longer needed.

PR #398, opened 2023-07-03
Update assets.blend

TheMakerTales:
Adding Fill Mesh, Fill Curve & Linear Array


hlorus:
Looks good to me! Just wondering if the fill groups could be merged and check the type of geometry implicitly?

PR #421, opened 2023-11-13
Treat has_coincident like a function

bonjorno7:
There's no way this was intentional.


hlorus:
Nice catch, thanks alot

PR #425, opened 2023-11-19
Improve auto constraints

bonjorno7:
By using a context manager which prevents you from adding constraints that stop the sketch from solving.

### DIFF
--- a/operators/utilities.py
+++ b/operators/utilities.py
@@ -208,7 +208,12 @@ def select_target_ob(context, sketch):
 
 
 @contextmanager
-def safe_constraints(context: Context, sketch: SlvsSketch = None, constraints: SlvsConstraints = None):
+def safe_constraints(
+    context: Context,
+    sketch: SlvsSketch = None,
+    constraints: SlvsConstraints = None,
+    remove_only_failed=False,
+):
     sketch = sketch or context.scene.sketcher.active_sketch
     constraints = constraints or context.scene.sketcher.constraints
 
@@ -226,7 +231,7 @@ def safe_constraints(context: Context, sketch: SlvsSketch = None, constraints: S
             new: set[GenericConstraint] = set(constraints.all) - old
 
             for constraint in new:
-                if constraint.failed:
+                if constraint.failed or not remove_only_failed:
                     constraints.remove(constraint)
 
             sketch.solve(context, update_entities=False)
--- a/operators/utilities.py
+++ b/operators/utilities.py
@@ -1,4 +1,5 @@
 import logging
+from contextlib import contextmanager
 
 import bpy
 from bpy.types import Context, Operator
@@ -7,6 +8,9 @@ from mathutils import Matrix
 from .. import global_data
 from ..declarations import GizmoGroups, WorkSpaceTools
 from ..converters import update_convertor_geometry
+from ..model.group_constraints import SlvsConstraints
+from ..model.sketch import SlvsSketch
+from ..solver import solve_system
 from ..utilities.preferences import use_experimental, get_prefs
 from ..utilities.data_handling import entities_3d
 
@@ -201,3 +205,25 @@ def select_target_ob(context, sketch):
     if target_ob.name in context.view_layer.objects:
         target_ob.select_set(True)
         context.view_layer.objects.active = target_ob
+
+
+@contextmanager
+def safe_constraints(context: Context, sketch: SlvsSketch = None, constraints: SlvsConstraints = None):
+    sketch = sketch or context.scene.sketcher.active_sketch
+    constraints = constraints or context.scene.sketcher.constraints
+
+    solve = solve_system(context, sketch)
+    count = len(list(constraints.all))
+
+    try:
+        yield constraints
+
+    finally:
+        # Don't bother if the sketch can't be solved.
+        if not solve:
+            return
+
+        # Remove the newest constraint until the sketch solves.
+        new = list(constraints.all)[count:]
+        while not solve_system(context, sketch):
+            constraints.remove(new.pop())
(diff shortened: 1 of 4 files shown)


bonjorno7:
The linter seems to dislike long lines.
I didn't think they would be an issue, because the project has lots of them already.
I suppose I can split things across more lines if necessary.

hlorus:
Great stuff! I see one major problem tho, since constraints.all is not sorted by creation date the contextmanager might remove constraints that weren't added in its context which is bad IMO.
Maybe we could record a list of added constraints or just check how many have been added per type.

Furthermore checking if the system solves shouldn't change entities. Maybe we can add an option to the solver which doesn't apply changes.

operators/utilities.py
```diff
@@ -201,3 +205,25 @@ def select_target_ob(context, sketch):
...
         target_ob.select_set(True)
         context.view_layer.objects.active = target_ob
+
+
+@contextmanager
+def safe_constraints(context: Context, sketch: SlvsSketch = None, constraints: SlvsConstraints = None):
+    sketch = sketch or context.scene.sketcher.active_sketch
+    constraints = constraints or context.scene.sketcher.constraints
+
+    solve = solve_system(context, sketch)
```
hlorus, line 215: We can probably avoid the initial solve check and just get the value from sketch.solver_state

hlorus:
> The linter seems to dislike long lines.
> I didn't think they would be an issue, because the project has lots of them already.
> I suppose I can split things across more lines if necessary.

Don't worry about the linter too much, it's more of an aid.

operators/utilities.py
bonjorno7: Oh I wasn't aware of that, I'll use it instead of calling the function then yeah.

bonjorno7:
> Great stuff! I see one major problem tho, since constraints.all is not sorted by creation date the contextmanager might remove constraints that weren't added in its context which is bad IMO. Maybe we could record a list of added constraints or just check how many have been added per type.

I assumed that it'd be sorted, my bad. How would you go about getting a list that *is* sorted?
Alternatively I can use a set and subtract them to get the new constraints; though they would be in arbitrary order.

> Furthermore checking if the system solves shouldn't change entities. Maybe we can add an option to the solver which doesn't apply changes.

I'm not sure what you mean by this (probably because I don't know much about this code base).

bonjorno7:
It now only removes constraints that are actually new.
If the sketch fails, it removes all of the new constraints.
Ideally it'd find out which constraints specifically are conflicting and remove those, but I don't know how to do that at this time.

hlorus:
> It now only removes constraints that are actually new.
> If the sketch fails, it removes all of the new constraints.
> Ideally it'd find out which constraints specifically are conflicting and remove those, but I don't know how to do that at this time.

I think that's a fair tradeoff and probably easy for the user to understand as the tool either creates constraints or not.
Just to note: the solver reports which constraints are failing, you can check that through constraint.failed. We could potentially try to first remove the failed, newly-created constraints however i'm not sure how consistent that behavior would be.

hlorus:
> > Furthermore checking if the system solves shouldn't change entities. Maybe we can add an option to the solver which doesn't apply changes.
> 
> I'm not sure what you mean by this (probably because I don't know much about this code base).
> 

The solver simply changes the position of geometric entities based on the constraints. We don't want that to happen always, usually only from the fini method of operators. With this PR the solver is triggered from the "main" method of the line_2d operator which runs every frame and causes the first point of the line to follow the vertical/horizontal linewhich is bad.
I think it should be enough to add a new option to the solve method in solver.py, something like update_entities=True or similar.

bonjorno7:
> The solver simply changes the position of geometric entities based on the constraints. We don't want that to happen always, usually only from the fini method of operators. With this PR the solver is triggered from the "main" method of the line_2d operator which runs every frame and causes the first point of the line to follow the vertical/horizontal linewhich is bad. I think it should be enough to add a new option to the solve method in solver.py, something like update_entities=True or similar.

Ah alright, makes sense.
I've added the update_entities option (True by default), so safe_constraints no longer updates entities.

> I think that's a fair tradeoff and probably easy for the user to understand as the tool either creates constraints or not.
Just to note: the solver reports which constraints are failing, you can check that through constraint.failed. We could potentially try to first remove the failed, newly-created constraints however i'm not sure how consistent that behavior would be.

I made it remove only the constraints where failed=True; though it might make sense to make this an option.
I feel like if you're adding multiple constraints inside the same safe_constraints block, it's probably because they rely on each other.
So it might make more sense to remove all of the constraints in that case, rather than only the failed ones.

bonjorno7:
> I made it remove only the constraints where failed=True; though it might make sense to make this an option.
I feel like if you're adding multiple constraints inside the same safe_constraints block, it's probably because they rely on each other.
So it might make more sense to remove all of the constraints in that case, rather than only the failed ones.

Added the option, so by default it removes all new constraints if any of them failed; if enabled it only removes failed constraints and keeps the successful ones.

bonjorno7:
Oh and would it be possible to set the linter to 120 line length?
That'd be consistent with the actual code base, so it would stop failing every time.

hlorus:
Looks good so far. Do you plan to change the other tools that add constraints as well with this PR?

bonjorno7:
If you think the context manager is good now, I can start adding the other operators yeah.

hlorus:
> If you think the context manager is good now, I can start adding the other operators yeah.

It looks good to me so far👍🏼

bonjorno7:
I could use some help coming up with cases where auto constraints fail actually.
It's easy enough for line_2d, but other than that I'm not sure.

hlorus:
I think it's mainly a problem with some advanced tool like bevel/trim. If there are no cases where it fails with the simple tools we can just leave them as they are.

bonjorno7:
I forgot there are other non-okay states that are not inconsistent, so the code doesn't work as it should in all cases.
I'm working on fixing that at the moment.

hlorus:
This has been resolved by #645, closing.

PR #441, opened 2024-01-18
Fix tangent typo

dcousens:
The UI description for the tangent operator shows `tagent` instead of `tangent`.
I think this is where that typo is coming from...


hlorus:
Thanks for the fix!

PR #443, opened 2024-01-24
Update Blender scripts-path config in installation.md

DNX:
Update Blender scripts-path config in installation.md to match the right location in Preferences as in Blender >= 4.0

image


hlorus:
Thanks

PR #457, opened 2024-04-08
Check user's Blender version against minimum required Blender version.

colossatr0n:
I had attempted to resolve some errors I was getting when activating the add-on and after some time realized that my Blender version was too low.

Added a version check that throws an exception if the user's Blender version is too low. This check happens as soon as possible in the code execution to prevent users from attempting to solve downstream issues.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,15 +1,6 @@
 import logging
 
-from bpy.app import background
-
-from . import global_data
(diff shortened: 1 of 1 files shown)


hlorus:
Thanks for the PR, merged with a change to take the minimal version from blender_manifest.toml which will be used with blender 4.2 and beyond.

PR #460, opened 2024-04-16
Mesh Extrude, Boolean Convert type

SpectralVectors:
## Mesh Extrude
MeshExtrude
Added an option to 'Extrude' a 3rd dimension to sketches when converting to meshes.  
This adds a Solidify modifier automatically and fills in the 'Depth' and 'Offset' from sketch properties `extrude_depth` and `extrude_offset` It first checks for and adjusts the existing 'Extrude Sketch' Solidify modifier, and creates one if it doesn't exist.

## Boolean Convert type
BooleanConvert
Added a 4th conversion mode to the `update_convertor_geometry` function - `Boolean` 
This will convert to mesh, and allows for the new 'Extrude', but also sets the viewport display to 'Bounds' and allows you to specify the target object you wish to cut, automatically applying a Boolean modifier to the target, and selecting your current sketch as the cutter object.

### DIFF
--- a/converters.py
+++ b/converters.py
@@ -96,7 +96,7 @@ def mesh_from_temporary(mesh: Mesh, name: str, existing_mesh: Union[bool, None] 
 
 
 def _cleanup_data(sketch, mode: str):
-    if sketch.target_object and mode != "MESH":
+    if sketch.target_object and mode not in ["MESH", "BOOLEAN"]:
         sketch.target_object.sketch_index = -1
         bpy.data.objects.remove(sketch.target_object, do_unlink=True)
         sketch.target_object = None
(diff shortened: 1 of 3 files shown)


hlorus:
Hi @SpectralVectors 

I apologize for the huge delay in responding to your pull request. Your contribution looks great, but unfortunately, it conflicts with the solution we decided on, which is currently in experimental mode.

You can check out the experimental solution under "Node Tools" in the tools panel. To access it, please enable the "Show Debug Settings" option. Since such tools aren't the main topic of the addon we wanted a system that is somehow flexible and potentially allows users to customize the behavior.

We'd definitely appreciate contributions that follow this implementation.

Thank you for your understanding and for your valuable contribution!

SpectralVectors:
All good, I'll have to check out the latest version.
Thanks for the addon!

Issue #459, opened 2024-04-12
Cannot set Distance constraint to horizontal or vertical for lines [BUG]


TimStrauven:
### Description

When adding a distance constraint on a line, it will be placed along the parallel of that line. In the options of the constraint it displays horizontal and vertical, but clicking it results in an error in "model/distance.py" line 72. 
This line is: distance = _get_aligned_distance(self.entity1, self.entity2, alignment)
While on a line distance constraint entity2 is "None".
This then triggers an error in _get_aligned_distance which is clearly created to deal with points:
```Python
def _get_aligned_distance(p_1, p_2, alignment):
    if alignment == "HORIZONTAL":
        return abs(p_2.co.x - p_1.co.x)
    if alignment == "VERTICAL":
        return abs(p_2.co.y - p_1.co.y)
    return (p_2.co - p_1.co).length
```

As it was designed for points in the first place my suggested solution would be to never place distance constraints on lines, but always detect the endpoints and always use these points instead.
This can be done in  "operators/add_distance.py" in the main method by checking if self.entity1 is a 2D line and self.entity2 is None. Then getting the endpoints of the line and feeding those into the add_distance function.

### Addon Version

0.27.3

### Blender Version

4.1

### What platform are you running on?

Linux

TimStrauven:
It took some effort to set entity1 and 2 since there are not setters directly, however after figuring out how the stateful operators were implemented I could set them this way. This solution works well, also with the other proposed change to check which constraint exists.

The changes are as explained above in "operators/add_distance.py"
extra imports:
```Python
from ..model.line_2d import SlvsLine2D
from ..model.point_2d import SlvsPoint2D
```

and updated main:
´´´Python
    def main(self, context):
        if type(self.entity1) == SlvsLine2D and self.entity2 == None:
            dependencies = self.entity1.dependencies()
            if type(dependencies[0]) == SlvsPoint2D and type(dependencies[1]) == SlvsPoint2D:
                for i in range(0,2):
                    self.state_index = i # set index to access i state_data
                    self.state_data["hovered"] = -1
                    self.state_data["type"] = type(dependencies[i])
                    self.state_data["is_existing_entity"] = True
                    self.state_data["entity_index"] = dependencies[i].slvs_index      

        if not self.exists(context, SlvsDistance):
            self.target = context.scene.sketcher.constraints.add_distance(
                self.entity1,
                self.entity2,
                sketch=self.sketch,
                init=not self.initialized,
[...]

Issue #430, opened 2023-12-09
[BUG] CAD Sketcher allows us to apply the hor/vert constraints multiple times to an element


raindropsfromsky:
### Contact Details

raindrops.fromsky@gmail.com

### Description

CAD sketcher allows us to apply the Hor/Vert constraints multiple times to an element, as shown below.

![image](https://github.com/hlorus/CAD_Sketcher/assets/9047168/20c24fa5-0a21-4aeb-8b7b-b342eb319f3a)

**Desired:** Allow the hor/Vert constraints to be applied only once. If I apply the same constraint once again, CAD Sketcher should ignore that input.

### Addon Version

0.27.3

### Blender Version

4.0 and 4.1

### What platform are you running on?

Windows

TimStrauven:
Hi hlorus, 

Since I was already found the solution for the other constraint issue i took a look at this one as well.

The reason here is in the file "operators/add_geometric_constraints.py", the main function of "VIEW3D_OT_slvs_add_horizontal" does indeed not check if the constraint already exists.

I solved it as follows, with two extra imports on the file:

´´´Python
from ..model.horizontal import SlvsHorizontal
from ..model.line_2d import SlvsLine2D

class VIEW3D_OT_slvs_add_horizontal(Operator, GenericConstraintOp):
    """Add a horizontal constraint"""

    bl_idname = Operators.AddHorizontal
    bl_label = "Horizontal"
    bl_options = {"UNDO", "REGISTER"}

    type = "HORIZONTAL"

    def main(self, context):    
        # check if the constraint already existed on the selected line or points
        hor_constr_exists = False
        for c in context.scene.sketcher.constraints.all:
            if self.entity1 in c.dependencies() and type(c) == SlvsHorizontal:
                if type(self.entity1) == SlvsLine2D:
                    hor_constr_exists = True
                elif self.entity2 in c.dependencies():
                    hor_constr_exists = True

        if not hor_constr_exists:
            self.target = context.scene.sketcher.constraints.add_horizontal(
                    self.entity1,
                    entity2=self.entity2,
                    sketch=self.sketch,
                )
        
        return super().main(context)
```

I was still looking to get the constraints from the active sketch only instead of all to save some items from the loop, but could not find it.

It works as intended for all tests i could come up with, please let me know any comments or thoughts. 
If everything looks ok, I'll do the same for vertical constraints and create a PR.

Regards,
T

hlorus:
[...]

PR #461, opened 2024-04-17
Checks for duplicate constraints.

TimStrauven:
On constraint creation checks are added to see if it exists on the selected entity or entitiies. The exists method was added to GenericConstraintOp for this. 
Line distances have been replaced for point to point distances.

linked to:
[BUG] CAD Sketcher allows us to apply the hor/vert constraints multiple times to an element #430 
 Cannot set Distance constraint to horizontal or vertical for lines [BUG] #459


operators/add_distance.py
```diff
@@ -33,13 +36,27 @@ class VIEW3D_OT_slvs_add_distance(Operator, GenericConstraintOp):
...
-            **self.get_settings(),
-        )
+        if isinstance(self.entity1, SlvsLine2D) and self.entity2 is None:
+            dependencies = self.entity1.dependencies()
+            if (isinstance(dependencies[0], SlvsPoint2D) and 
+                    isinstance(dependencies[1], SlvsPoint2D)):
+                # for loop changes the values of self.entity1 and self.entity2
+                # from a line entity to its endpoints
+                for i in range(0, 2):
+                    self.state_index = i  # set index to access i state_data
```
hlorus, line 46: This is problematic, state_index is used to specify the currently active state, to access the data of a specific state you can use the following:
data = self.get_state_data(index)

operators/base_constraint.py
```diff
@@ -124,3 +130,33 @@ def draw(self, context: Context):
...
+                max_distance_dof = 1
+        else:
+            new_dependencies = [i for i in [self.entity1, self.sketch] if i is not None]
+            max_distance_dof = 1
+
+        distance_dof = 0
+        for c in context.scene.sketcher.constraints.all:
+            if isinstance(c, constraint_type):
+                if set(c.dependencies()) == set(new_dependencies):
+                    if constraint_type == SlvsDistance:
```
hlorus, line 154: It seems quite odd to have specific type checks in the base class, this should be done in the relevant sub-class (e.g. VIEW3D_OT_slvs_add_distance).

hlorus (changes requested):
Looks good to me besides the two comments.

hlorus:
> Line distances have been replaced for point to point distances.
> 
> linked to: #430 Cannot set Distance constraint to horizontal or vertical for lines [BUG] #459

Note: In order to solve 430 we'll have to migrate old distance constraints to also contrain point to point.

operators/add_distance.py
TimStrauven: Ok, does this look better?
```python
                for i in range(0, 2):
                    state_data = self.get_state_data(i)
                    state_data["hovered"] = -1
                    state_data["type"] = type(dependencies[i])
                    state_data["is_existing_entity"] = True
                    state_data["entity_index"] = dependencies[i].slvs_index
                self.next_state(context)
```
We still need to make an extra call to self.next_state, because otherwise it will not detect the end of the selection process and expect a second entity from the user.

operators/base_constraint.py
TimStrauven: We can add the whole part to VIEW3D_OT_slvs_add_distance the main would then look like this:
```python
 def main(self, context):
        if isinstance(self.entity1, SlvsLine2D) and self.entity2 is None:
            dependencies = self.entity1.dependencies()
            if (isinstance(dependencies[0], SlvsPoint2D) and 
                    isinstance(dependencies[1], SlvsPoint2D)):
                # for loop changes the values of self.entity1 and self.entity2
                # from a line entity to its endpoints
                for i in range(0, 2):
                    state_data = self.get_state_data(i)
                    state_data["hovered"] = -1
                    state_data["type"] = type(dependencies[i])
                    state_data["is_existing_entity"] = True
                    state_data["entity_index"] = dependencies[i].slvs_index
                self.next_state(context)

        if (isinstance(self.entity1, (SlvsPoint3D, SlvsLine3D)) or 
                isinstance(self.entity2, (SlvsPoint3D, SlvsLine3D))):
            max_constraints = 3
        elif ((isinstance(self.entity1, SlvsLine2D) and self.entity2 is None) or 
                isinstance(self.entity1, SlvsPoint2D) and isinstance(self.entity2, SlvsPoint2D)):
            max_constraints = 2
        else:
            max_constraints = 1

        if not self.exists(context, SlvsDistance, max_constraints):
            self.target = context.scene.sketcher.constraints.add_distance(
                self.entity1,
                self.entity2,
                sketch=self.sketch,
                init=not self.initialized,
                **self.get_settings(),
            )
        return super().main(context)
```

and the exists in the base class:
```python
    def exists(self, context, constraint_type=None, max_constraints=1) -> bool:
        if hasattr(self, "entity2"):
            new_dependencies = [i for i in [self.entity1, self.entity2, self.sketch] if i is not None]
        else:
            new_dependencies = [i for i in [self.entity1, self.sketch] if i is not None]

        constraint_counter = 0
        for c in context.scene.sketcher.constraints.all:
            if isinstance(c, constraint_type):
                if set(c.dependencies()) == set(new_dependencies):
                    constraint_counter += 1
                    if constraint_counter >= max_constraints:
                        return True

        return False
```

Does this all look better? It does to me...

operators/add_distance.py
hlorus: Usually the StatefulOperator class handles the logic to step through the states. Implementation classes shouldn't change that logic and only care about their specific functionality. This is however a local problem with small risk, so i'm ok with the solution if it works.

operators/base_constraint.py
hlorus: Yes that looks better

hlorus:
To migrate existing constraints:
- Bump the addon version (e.g. 0.27.4) in __init__.py bl_info and blender_mainfest.toml
- Add a check for files older than that new version in versioning.py
- Loop through constraints of type distance and replace line entity with the two points of the line

TimStrauven:
> To migrate existing constraints:
> 
>     * Bump the addon version (e.g. 0.27.4) in **init**.py bl_info and blender_mainfest.toml
> 
>     * Add a check for files older than that new version in versioning.py
> 
>     * Loop through constraints of type distance and replace line entity with the two points of the line

Yes I figured this out yesterday, but while looking trough it I noticed the versioning system is broken for blender before 4.20 because of the changes made in the last month. I'll create another bug report...

TimStrauven:
> To migrate existing constraints:
> 
>     * Bump the addon version (e.g. 0.27.4) in **init**.py bl_info and blender_mainfest.toml
> 
>     * Add a check for files older than that new version in versioning.py
> 
>     * Loop through constraints of type distance and replace line entity with the two points of the line

Think this is all. I added quite some checks into versioning to be really sure to select only the right ones (since e.g. angle constraints are also in dimensional).

Let me know if anything doesn't look right.

versioning.py
```diff
@@ -100,5 +100,22 @@ def do_versioning(self):
                     )
                 )
                 c.is_reference = True
+        if version < (0, 27, 4):
```
hlorus, line 103: A few things to note:
- Imports can be avoided, constraints can be checked by constraint.type == "DISTANCE", entities by using the "is_xxx" methods (e.g. entity.is_line() and entity.is_2d()).
- In general it's nice to avoid deep nesting by inverting the if checks in a loop, just do:
    if c.type != "DISTANCE": continue

hlorus (approved):

versioning.py
TimStrauven: I am not sure about only checking constraint.type == "DISTANCE", entities by using the "is_xxx" methods (e.g. entity.is_line() and entity.is_2d()). 
This bit of code can change quite a lot in a users drawing. I went from the assumption that you cannot know what weird stuff happened in their drawing and everything needs to be checked first. I think this is the only way to be sure? I also added an extra length check on line_dependencies just to cover such cases. 

Correct about the nested if statements, I pushed an update.

hlorus: I'm not saying you should remove checks, just  to replace them with the corresponding method calls. This can help with readability and make the imports unnecessary. 

Anyway this was just a note, i'm fine with either way. :)

TimStrauven:
> This is ok to go in, feel free to merge yourself when you're ready. I've added you as a collaborator to the project.

Thanks, but i've declined the request for now, maybe at a later point when I know all details of the project better.

Issue #389, opened 2023-06-08
Bevel Tool Supporting Lines For Constrained Points


TheMakerTales:
Currently, when adding a bevel to a point that already has a dimensional constant it's behavior is this:

![Capture1](https://github.com/hlorus/CAD_Sketcher/assets/103122109/821bc3a8-fb82-472f-ae51-0a3871efcdbb)

Let's add some supporting construction like creation for point that are beveled that are already constrained in some way like this:

![Capture](https://github.com/hlorus/CAD_Sketcher/assets/103122109/a8998918-2f0e-4984-bbb7-69526b879658)

I'm unsure if this should be "automatic" behavior, toggleable or modifier key initiated i.e. pressing shift while using the bevel tool

TheMakerTales:
Just to be clear the resulting construction lines should be parallel to the original line thus supporting lines in any orientation.

TimStrauven:
This behavior is linked to:
 Cannot set Distance constraint to horizontal or vertical for lines [BUG] #459 
It should be half solved if the PR on it is accepted:
 Checks for duplicate constraints. #461 

The reason it happened is also because the distance constraint is set onto the line instead of its endpoints.
New behavior after merge of that PR would look like this:
![image](https://github.com/hlorus/CAD_Sketcher/assets/96440176/ee9a90fe-34ee-452f-9bec-01f89b6e8c06)

So if the bevel tool would then add a coincident from the point to each line it should be solved as well I think.

hlorus:
Good catch, can you open a PR?

TimStrauven:
If you also want the lines to be added, this seems to work well  and I think it fully covers the initial request:

```Python
        # Remove original point if not referenced
        if not is_entity_referenced(point, context):
            context.scene.sketcher.entities.remove(point.slvs_index)
        else:
            ssc.add_coincident(point, seg1, sketch)
            ssc.add_coincident(point, seg2, sketch)

            # add reference construction lines
            sse = context.scene.sketcher.entities
            for i in range(0, 2):
                if isinstance(self.connected[i], SlvsLine2D):
                    target = sse.add_line_2d(point, self.points[i], sketch)
                    target.construction = True
                elif isinstance(self.connected[i], SlvsArc):
                    target = sse.add_arc(sketch.wp.nm, self.connected[i].ct, self.points[i], point, sketch)
                    target.construction = True
                    if target.angle > HALF_TURN:
                        target.invert_direction = True
´´´

[...]

PR #462, opened 2024-04-20
Updates for bevel tool

TimStrauven:
Adds coincident constraints if startpoint has an attached reference. Also adds construction lines/arcs to the reference startpoint.

This fixes:
 Bevel Tool Supporting Lines For Constrained Points #389 
But is dependent on feature "Line distances have been replaced for point to point distances." in PR:
 Checks for duplicate constraints. #461


hlorus:
This works well! Only thing i found is that when there is a construction arc segment added the coincident constraint is redundant.

hlorus (changes requested):
See comment above

TimStrauven:
> This works well! Only thing i found is that when there is a construction arc segment added the coincident constraint is redundant.

Indeed, updated to only set them on the lines.

Issue #162, opened 2022-06-04
[BUG] Construction circles not visible


BrCadfael:
### Contact Details

policebox@charter.net

### Description

When I set the Construction check on a circle, it disappeared.  Turning off the Construction check restores the circle.  I verified that the visible check was set the whole time.  My guess is that it doesn't know how to draw a construction circle.

### Addon Version

0.22.6

### Blender Version

3.1.2

### What platform are you running on?

Windows

BrCadfael:
Additional information.  The circles do appear if you zoom into them.  The problem appears to be connected to the size of the circle on the screen.  A large construction circle will be visible, a smaller one may not be.  There is a transition point where the circle "materializes" as you zoom in and "dematerializes" when you zoom out.  On my screen that happens when the circle is about a centimeter across.  The issue doesn't appear to affect lines at all.  Construction lines are always visible.

hlorus:
Looks like a glitch with the opengl dashed shader. I can reproduce this but it's usually not a problem for the scale i'm working with.

amrsoll:
For some context
![image](https://user-images.githubusercontent.com/24782867/183736182-6c19b84a-444f-4427-9fbc-21bfedd16b45.png)

![image](https://user-images.githubusercontent.com/24782867/183736251-c422f95f-435f-4b84-998b-e58c35f89e6a.png)

![image](https://user-images.githubusercontent.com/24782867/183736307-82f96117-55dd-4dee-a507-42d8a564d9cb.png)

jvenborg:
Yep, that has become an issue for me as well. Solid lines in circles are all fine, but construction lines will become invisible. In my current sketch, the diameters of the circles are 2 and 5 mm, respectively. Version 0.25.0.

Miguel-Rodrigues:
Same thing happens here:


![Image](https://github.com/hlorus/CAD_Sketcher/assets/10254384/a295209b-1362-4334-9962-fd6cc2b8c32a)


![Image](https://github.com/hlorus/CAD_Sketcher/assets/10254384/1c1ca8b2-6644-450f-98d0-6f0c2c1217dc)

Windows 10
Blender 3.5.1
CAD Sketcher 0.27.1

PR #472, opened 2024-07-29
Start dashes where line intersects the edge of the screen

AnyOldName3:
This improves on the problem where circles and other things made of lots of short lines become invisible, but the dash spacing still isn't brilliant. It helps #162, but I'm not sure I'd necessarily consider it fully fixed, so I'll leave it up to you whether you want to award the bounty.

Before:
image

After:
image

Alternatively, as an imgsli: https://imgsli.com/MjgyOTMw

I did attempt a better solution (you can see it here https://github.com/AnyOldName3/CAD_Sketcher/tree/sad-overcomplicated-dashhes) but it didn't end up actually looking any better, and was more complicated. Apparently, screen-space line stippling that continues nicely along separate segments in the same line strip isn't something that's easy to do, with the typical solutions being to stick to OpenGL 2.1, which does it in hardware, to use a Vulkan extension that does it in hardware, or to give up.

### DIFF
--- a/shaders.py
+++ b/shaders.py
@@ -21,12 +21,24 @@ class Shaders:
            gl_Position = ModelViewProjectionMatrix * vec4(pos.xyz, 1.0f);
 
            vec2 ssPos = vec2(gl_Position.xy / gl_Position.w);
-           stipple_start = stipple_pos = ssPos;
+           segment_start = stipple_pos = ssPos;
         }
     """
     base_fragment_shader_3d = """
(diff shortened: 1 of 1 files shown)


hlorus:
Hey sorry, i've overlooked your PR. This is a great improvement, i agree that it's not perfect but at least the entities are visible with this PR applied. Feel free to collect the bounty over at: https://opencollective.com/cadsketcher

Issue #426, opened 2023-11-30
[BUG] The "Add workplane" does not work consistently with "Add workplane on mesh face"


raindropsfromsky:
### Contact Details

raindrops.fromsky@gmail.com

### Description

When we click on the "Add workplane" button, CAD Sketcher shows us three intersecting planes, and when we hover the mouse on any plane, it gets highlighted. We click LMB on any plane to start working in it. This is intuitive.

Well, the "Add workplane on mesh face" button could have behaved in very similar fashion: When we hover the mouse on any face of an existing geometry, a plane should have appeared on that face. The user could have clicked LMB to confirm it. 

Instead, CAD Sketcher does not show us any plane at all. We have to click on the face, and then click RMB. Only after that does a plane appears on the face. It is not intuitive for a user to right-click to complete the operation, because the same action actually **cancels** the current operation in Blender!

These two workflows are totally different and also inconsistent with how Blender itself works.

Therefore, the workflow of CAD Sketcher should be as outlined above.

### Addon Version

0.27-3

### Blender Version

4.1 (actually, BforArtists 4.0.1)

### What platform are you running on?

Windows

hlorus:
The separate "Add Workplane" / "Add Workplane on Face" operators no longer exist — workplane selection is now part of Add Sketch, where you pick an origin plane or hover a mesh face (with a live preview) as the sketch base. Please retest on the latest prerelease build. Reopen if it persists.

PR #476, opened 2024-08-26
Fixed issue #426

MagicBOTAlex:
This might be a patch for the problem, but it works.
I noticed that Blender doesn't update the UI after creating a new workplane.

# Before
blender_vjybzlzsDG
I noticed that it creates the new workplanes, but doesn't update the changes.

# After
blender_iYdK64yN76
This is when I force the update after the workplane creation

Windows
Blender 4.0
CAD_Sketcher: latest
(I don't want the bounty. Too much for such a small change)

Edit: Hmm, I just thought about it. It might cause some performance problems for bigger projects. But at least I found the problem.

With that said, I'm not sure if it's avoidable to not re-draw.

### DIFF
--- a/operators/add_workplane.py
+++ b/operators/add_workplane.py
@@ -137,7 +137,7 @@ class View3D_OT_slvs_add_workplane_face(Operator, Operator3d):
 
         self.target = sse.add_workplane(origin, nm)
         ignore_hover(self.target)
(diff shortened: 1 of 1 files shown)


hlorus:
Thanks for the PR, this is looking good. The only note i have is that you'll likely don't need to loop through all areas but can just access the one that is set in the context, e.g.:
`context.area.tag_redraw()`

MagicBOTAlex:
Ohh, that's what "context" was. But it seems to be working. 👍

Issue #36, opened 2022-04-25
Project lines form objects [REQUEST]


gooch12013:
It would be nice to be able to project lines from objects.  this would be extremely helpful when building more complex objects

mullcom:
I agreed. When i get stuck in remaking or build complexity things this is a must to have to get a why to find a patch forward.

franklincg:
I'd like to take this bounty if it is still available. Estimated dev time: ~2–4 days including implementation, regression tests, and Blender validation.

I want the Bounty - Medium ($250) for completion.

I reviewed the existing projection work in PRs #481 and #518. Both are still open but currently non-mergeable/stalled, and the original #481 author explicitly encouraged a takeover. I would implement against current main, keeping projection as a separate operation and preserving a live reference to source geometry so projected entities update when the source changes, as requested in the maintainer feedback.

Please assign/lock the bounty to me if you’re happy for me to take it over.

PR #481, opened 2024-09-01
Line projection implimentation

MagicBOTAlex:
This is for issue #36

There are still a few problems/TODOs:
- [ ] Auto align view with sketch after creation
- [ ] Make it auto enter sketch
- [ ] Make the properties of the blender operator work correctly
- [ ] Maybe a little more clean-up in the code

The commits should maybe be squashed into a single one.
I really didn't think I would be able to implement this, so I just messed around. But here we are... It's working! :)

# Here's an GIFs of the projection in action:
ezgif-5-ce15cd38bc

###### If you find any problems with this code, please let me know. This is my first BIG github commit


hlorus:
Hey there, this is an interesting PR, thanks!
I haven't looked into the details but some notes regarding the fundamentals:

- It appears that the projected entities are just static, fixed entities which won't be updated when the base object is edited (Same as the workplane on mesh face...). If we add projection it would be great to also look into entities that keep a reference to the projected geometry. I was investigating this quite some time ago but unfortunately didn't continue as it wasn't a priority. A good place to look at is the following commit which we might be able to use without a big hassle: https://github.com/hlorus/CAD_Sketcher/commit/c55f84da6fa94b27cc0db0c7378be4b8a7e7dbb2
- It would be great to move the code out of the add_sketch.py file.
- Projection should be accessible as a separate operation. This also helps testing.

MagicBOTAlex:
1. I'll have a look at that commit, and I'd have to think about it for a bit.
2. Sure, I'll do that at some point
3. What are you thinking? Are we talking in sketch mode or in normal blender view mode? Would it be like in sketch mode, then project to workplane from any mesh?

Since, Uni my time will be much much more limited. I underestimated the severity of the assignments, lol.

hlorus:
> 3. What are you thinking? Are we talking in sketch mode or in normal blender view mode? Would it be like in sketch mode, then project to workplane from any mesh?
> 

I'd say ideally it should also work in 3d, so having a workspace tool would make sense.

MagicBOTAlex:
Hey, I'm sorry but I feel bad for not leaving a comment about my situation.
I have lost interest in this project as of now, so if someone else is willing to takeover, it's encouraged.
I may return to this in the distant future, when I lose access to Fusion 360 again. (Everytime I lose my access I re-visit this)

But currently, University is soo time consuming, and I have started on other projects.

To the next person taking over my progress (if there's someone):
- My code should have comments every now and then explaining it.
- "View3D_OT_slvs_add_sketch_face" is the main part of the code
- "ProjectFromMeshes" is the actual function doing the projection
- If there's any need for understanding my code, please write, I'll always be ready to help

hlorus:
Thanks for this. Project Geometry has since landed in #606, which covers this functionality, so closing as no longer needed.

PR #484, opened 2024-09-11
🐛 fix: improve exception for get_addon_version_tuple

IGLOU-EU:
When I started Blender, I didn't have the version of the add-on, but no error message. So I did a mini fix of the exception. 

The problem was that I didn't have the toml python lib. Is there anything I can do to indicate the need or autoinstall it ?

It's my first time with python, be kind :smile:


hlorus:
Thanks for the pr, looks good :)

PR #510, opened 2025-05-03
fix #439

### LINKED ISSUE #439: [BUG] Can't Force Update in Sketcher panel
### Contact Details

tluisrs@gmail.com

### Description

Working on this file I used to press the Force Update button to refresh all the meshes in the scene. Somehow I got CAD Sketcher into this state where force update no longer worker. I have removed the last sketch but I can't use force update in this file anymore.

image

### Addon Version

0.27.2

### Blender Version

3.5

Windows

rcarmo:
Using the provided sample file for #439, I managed to track it down to providing a "sensible" direction value when there might not be any.

### DIFF
--- a/converters.py
+++ b/converters.py
@@ -47,7 +47,12 @@ class BezierConverter(EntityWalker):
             last_index = len(path_segments) - 1
             index = 0
             for i, segment in enumerate(path_segments):
-                invert_direction = spline_path[1][i]
+                # Fix for missing direction information
+                if len(spline_path) < 2 or len(spline_path[1]) <= i:
+                    invert_direction = False
+                    logger.warning(f"Missing direction information for segment {i}, assuming False")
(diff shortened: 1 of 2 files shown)


hlorus:
Hey, thanks for looking into this. Do you also have a clue where that invariant comes from? There shouldn't really be a case where the direction of the entity is not set. The class EntityWalker in walker.py is responsible for constructing the this, possibly the mistake happens in the recursive walker function.

rcarmo:
No, but I can look into it further. At least right now this fixes the UX. I have gone down other rabbit holes in the meantime since I really needed to have selections working, but I've been finding some things in the debugger and making sure I can make progress...

hlorus:
Thanks for this. Issue #439 has since been resolved, so closing this as no longer needed.

PR #511, opened 2025-05-03
FIx #505

### LINKED ISSUE #505: [BUG] Can't select anything without using sketcher panel
### Contact Details

alistair.sutter@orange.fr

### Description

Hi,

I am using Cad Sketcher 0.27.6 on Ubuntu 24.04 with Blender 4.3.2.
I have also tested this with Cad Sketcher 0.27.2 and same behaviour.

When I click "Add Sketch" it shows me the 3 planes correctly. When I hover over them, nothing happens. When I click them they disappear and no sketch gets added.

Image

I have found a workaround and figured that the solvespace select is bugging out.
My workaround is to open the "Entities" dropdown in the sketcher panel and to select the wanted plane there. Otherwise it will not work by clicking on the plane in the viewport.

Image

Once selected I can start drawing entities. However, once again, I cannot select them in the viewport. I have to select them in the entity list and then apply constraints from there.
This time however, when I try to select an entity I get an error than may be helpful to find the problem:

Image

As you can imagine, this makes the cad sketcher workflow considerably slower than me just modeling vanilla in the viewport.

Please advise. I can send whatever logs/files you need.
Thanks.

### Addon Version

0.27.6

### Blender Version

4.3.2

Linux

rcarmo:
This is an attempt at fixing #505. I think that the offscreen buffer wasn't ready when hover detection was attempted and that selecting directly the plane wasn't initializing things properly. Either way, I now cannot reproduce the issue:

### DIFF
--- a/draw_handler.py
+++ b/draw_handler.py
@@ -15,13 +15,26 @@ logger = logging.getLogger(__name__)
 def draw_selection_buffer(context: Context):
     """Draw elements offscreen"""
     region = context.region
+    
+    # Only create a new offscreen if it doesn't exist or dimensions have changed
+    if (global_data.offscreen is None or 
+            global_data.offscreen.width != region.width or 
+            global_data.offscreen.height != region.height):
+        # Properly free the old offscreen before creating a new one
+        if global_data.offscreen is not None:
+            global_data.offscreen.free()
+            
+        try:
+            # create offscreen
+            width, height = region.width, region.height
+            offscreen = global_data.offscreen = gpu.types.GPUOffScreen(width, height)
+            logger.debug(f"Created new offscreen buffer: {width}x{height}")
+        except Exception as e:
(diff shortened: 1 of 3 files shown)


hlorus:
The changes look good to me. I cannot reproduce the initial issue so it's difficult to test, have you been able to reproduce the issue described in #505 ?

rcarmo:
I think this one is a red herring, to be honest. I was able to reproduce it before the "fix" but then I started looking at the entire selection mechanism and using the debugger in Blender 4.4 and did #513, which I think fixes this and a lot more indirectly. 

I'm running the plugin with all my recent PRs applied and selections work well for me.

hlorus:
Ok, should this PR still be merged or dropped in favor of #513?

rcarmo:
I say we drop it.

PR #512, opened 2025-05-03
Fix #489

### LINKED ISSUE #489: [BUG] Extra edges and vertices when converting to filled mesh
### Contact Details

deng@randomsample.de

### Description

Steps to reproduce:

* Create the following simple sketch:

cads_1

The exact dimensions do not matter, but make sure to have an equal constraint at the left and right vertical lines I marked with "A" and "B". Then set "convert type" to "mesh" and select "Fill Shape". This is the result:

cads_2

Note the additional horizontal edge and vertices I marked with the red circle. This should not be there and creates problems with further work (like when applying the screw modifier, this creates additional faces I don't want).

Note this only happens when converting to a filled mesh. If the mesh is not filled, it is fine. Likewise, a Bezier will work fine. Also, it only happens when the lines A and B are exactly equal, not matter how this is achieved. For instance, you can remove the equal constraint and use two distance constraints with the same length, it'll also happen, but just the slightest difference in distance will make this problem disappear.

Fortunately, someone has posted a workaround in Discord: you can disable "Fill Shape" and use a geometry node modifier like this to fill the mesh instead:

blender_3

Naturally, I'm wondering if maybe this can be implemented in CADS itself?

### Addon Version

latest

### Blender Version

4.2.2

Linux

rcarmo:
I fixed #489 by doing two passes over segments when converting and making sure things are collinear. It works for the simple initial case - I cannot get at the sample files that were linked later.

image

### DIFF
--- a/converters.py
+++ b/converters.py
@@ -81,9 +81,32 @@ def mesh_from_temporary(mesh: Mesh, name: str, existing_mesh: Union[bool, None] 
     bm = bmesh.new()
     bm.from_mesh(mesh)
 
+    # First pass: dissolve vertices that are nearly collinear (handles most cases)
     bmesh.ops.dissolve_limit(
         bm, angle_limit=math.radians(0.1), verts=bm.verts, edges=bm.edges
     )
+    
+    # Second pass: specifically target line segments with multiple unnecessary vertices
+    edges_to_dissolve = []
+    for edge in bm.edges:
+        if not edge.is_boundary:
+            continue
+        
+        # Check if vertex is connected to exactly 2 edges and is nearly collinear
+        for vert in (edge.verts[0], edge.verts[1]):
+            if len(vert.link_edges) == 2:
+                edges = list(vert.link_edges)
+                vec1 = edges[0].other_vert(vert).co - vert.co
+                vec2 = edges[1].other_vert(vert).co - vert.co
+                if vec1.length > 0 and vec2.length > 0:
+                    # Check if the two edges are nearly collinear
+                    angle = vec1.angle(vec2)
+                    if abs(angle - math.pi) < 0.01:  # Very close to 180 degrees
+                        edges_to_dissolve.append(vert)
+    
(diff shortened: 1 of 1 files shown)


rcarmo:
Well, I could try to check if I'm walking through a circle.

hlorus:
There's still an error when applying the steps described before:

ValueError: verts: found the same (BMVert) used multiple times
Error: Python: Traceback (most recent call last):
  File "/home/yoga/Dropbox/blend/extensions/CAD_Sketcher/operators/set_sketch.py", line 20, in execute
    return activate_sketch(context, self.index, self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoga/Dropbox/blend/extensions/CAD_Sketcher/operators/utilities.py", line 148, in activate_sketch
    update_convertor_geometry(context.scene, sketch=last_sketch)
  File "/home/yoga/Dropbox/blend/extensions/CAD_Sketcher/converters.py", line 238, in update_convertor_geometry
    mesh = mesh_from_temporary(
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/yoga/Dropbox/blend/extensions/CAD_Sketcher/converters.py", line 143, in mesh_from_temporary
    bmesh.ops.dissolve_verts(bm, verts=edges_to_dissolve)

Also, just as a note: Ideally we would get that is_curve flag from the entity, maybe through attributes.

rcarmo:
Hmm. I'm not getting that here. What Blender are you running, just for comparison?

hlorus:
Blender 4.4. This is a simple testfile where it happens when i leave the sketch:
[circle-resolution.zip](https://github.com/user-attachments/files/20032106/circle-resolution.zip)
(Just change the file ending from .zip to .blend)

hlorus:
Thanks for this. Issue #489 has since been resolved, so closing this as no longer needed.

PR #518, opened 2025-06-10
Feature: Add Project-Include Tool

dyami0123:
Building off of @MagicBOTAlex's great work in pr #481. addresses issue #36 

Major Changes:

added a "Project-Include" tool based off of the same tool within Fusion360. This tool allows the user to select a mesh object, or a face within a mesh object and project it onto the Workplane.

Will add example gif to this PR sometime soon

Tested a few options for "highlighting" geometry for easier selection. ended up needing to tweak the OperatorSubclass.evaluate_state() method to allow the desired behavior. I wanted the UX to be: when the tool is selected blender activates modal mode automatically and utilizes the highlighting logic present in the operator. After some light research, that didn't seem possible (though I'm fairly new to blender Addons), so i ended up with the following UX: the user selects the tool, clicks somewhere on the screen to activate modal mode, and the highlighting happens.

The highlighting itself is also somewhat janky, I couldn't get any custom GL code to run without crashing blender so I'm creating temporary meshes on screen with a highlight material (more visible in viewport shading). Definitely open to suggestions here, but this seemed to get the job done.

Additional Thoughts:

- made some changes to the underling logic within stateful_operator.logic to give the operator access to the event information. (if shift is held it changes the tool behavior)
- will revert the changes to keymaps, made those initially when getting started, and i dont think they make sense.
- i have an autoformatter on for imports, i can revert changes to existing files as needed.
- I'd love to connect the highlight color to the user settings eventually.

As mentioned in the previous PR, if we can find a way to connect the geometry to the sketch s.t. if the source geometry moves, so does the sketch projection then that would really make this work great. I'm not entirely sure how best to tackle that, so open to ideas there.

Additional Work:
- [ ] handle selections on un-supported objects without failure
- [ ] update icons and helper text
- [ ] create useful gif for docs
- [ ] Look into source geometry-sync


hlorus:
Thanks for this. Project Geometry has since landed in #606, which covers this functionality, so closing as no longer needed.

PR #543, opened 2026-03-11
Fix enum int/string handling in setprop (Blender 5.0) and _get_align callback

falken10vdl:
Blender's enum properties:
- RNA assignment via setattr expects string identifiers, 
- get/set property callbacks must return/receive integers.

bpy.py — setprop() (Blender 5.0 compat):
Blender 5.0 no longer accepts integers when assigning enum properties via setattr. Changed from `prop.enum_items[value].value (integer)` to `prop.enum_items[value].identifier (string)`. Added an `isinstance(value, int)` guard so string values pass through unchanged.

distance.py — SlvsDistance._get_align() (latent bug):
The get callback for the align EnumProperty was returning self.align_store directly — a string identifier — which Blender cannot interpret as an integer, causing:
`TypeError: 'str' object cannot be interpreted as an integer`
Fixed by converting the stored identifier back to its integer index via `bpyEnum(align_items, identifier=self.align_store).index`.

Cheers!

### DIFF
--- a/model/distance.py
+++ b/model/distance.py
@@ -71,7 +71,7 @@ class SlvsDistance(DimensionalConstraint, PropertyGroup):
     def _set_align(self, value: int):
         alignment = bpyEnum(align_items, value).identifier
         distance = _get_aligned_distance(self.entity1, self.entity2, alignment)
-        self.align_store = value
+        self.align_store = alignment
         self.value_store = distance
 
     def _get_align(self) -> int:
--- a/model/distance.py
+++ b/model/distance.py
@@ -52,6 +52,7 @@ align_items = [
     ("VERTICAL", "Vertical", "", 2),
 ]
 
+
 def _get_value(self):
     if self.is_reference:
         val = self.init_props(align=self.align)["value"]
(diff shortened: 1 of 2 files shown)


jrversteegh:
Incomplete? `_set_align`  should also be adjusted:
```
  File ".config/blender/5.1/extensions/user_default/CAD_Sketcher/model/distance.py", line 74, in _set_align
    self.align_store = value
    ^^^^^^^^^^^^^^^^
TypeError: bpy_struct: item.attr = val: expected a string enum, not int
```

kspiero (approved):

hlorus (commented):
As @jrversteegh pointed out _set_align is still broken (assigns int to align_store which now expects a string).

falken10vdl:
@hlorus  and @jrversteegh you are right! I made an update. Can you please take a look?
Thanks!

hlorus:
Looks good, thanks for the quick response!

falken10vdl:
@hlorus thanks to you guys! Outstanding project. Thanks!

PR #546, opened 2026-04-13
Fix sketch mesh creation sometimes creating degenerate faces, or faces with reversed normals

### LINKED ISSUE #373: [BUG] Trim tool causing problems with the generated mesh
### Contact Details

vermit25@yahoo.ca

### Description

If you use the trim tool to create a sketch the resulting mesh is quite the mess.
See video:
https://www.loom.com/share/2d03a43837294660bb52a51d0b439065

### Addon Version

0.27.1

### Blender Version

3.4.1

Windows

### LINKED ISSUE #447: [BUG] Generated geometry is incorrect and cannot be properly extruded using the solidify modifier
### Contact Details

_No response_

### Description

When creating a geometry using 2D arcs in a specific arrangement, the resulting bezier/mesh has some artifacts which make it difficult to continue working with the generated mesh.

Steps to reproduce:
1. Create new blender project
2. Sketch shapes using 2D arcs and lines (must have some complexity, e.g. as shown in the attached video)
3. Convert the sketch to a mesh
4. Apply the "Solidify" modifier and increase thickness to observe the artifacts

https://github.com/hlorus/CAD_Sketcher/assets/10092394/71886418-26cd-4328-9a53-75fad389443b

Download the model as shown in the video:
[broken_sketch_solidify.zip](https://github.com/hlorus/CAD_Sketcher/files/14271917/broken_sketch_solidify.zip)

(Apparently I don't get any logs from the plugin.)

PS: Huge shoutout to this project, it has already served me well in the past and I hope to continue my work with it!

### Addon Version

0.27.3

### Blender Version

4.0

Windows

### LINKED ISSUE #489: [BUG] Extra edges and vertices when converting to filled mesh

NeilHarbin0:
This fixes a couple of issues with mesh creation when exiting a sketch I have run into.

1. Fixes certain geometry creating faces that result in a normal vector of {0,0,0}, leading to solidify extruding incorrectly. Solution for this was dissolving the things with this incalculable normal vector result.

2. Fixes certain geometry creating faces that have vertexes spun in the wrong order, leading to a reversed normal vector compared to the rest. The solution for this was a bit more odd, checking the average normal of the faces, and then flipping ones opposite of that average direction.

Fixes #373
Fixes #447
Fixes #489

I used Claude AI to help me with all of this. I don't really write python, and certainly had no idea how blender plugins work or where in the repo the related code was for the meshing step. Fortunately there's not a lot of changed lines, so hopefully it is pretty understandable.

I tested these changes on the blend file from issue 447, the process from 373, the process from 489, and on my own blend file where I originally encountered the problem. It worked to fix all of them at least for using solidify.

Before:
image

After:
image

### DIFF
--- a/converters.py
+++ b/converters.py
@@ -5,6 +5,7 @@ from typing import Union
 import bpy
 import bmesh
 from bpy.types import Mesh, Scene, Object
+from mathutils import Vector
 
 from .utilities.bezier import set_handles
 from .utilities.walker import EntityWalker
@@ -84,6 +85,22 @@ def mesh_from_temporary(mesh: Mesh, name: str, existing_mesh: Union[bool, None] 
     bmesh.ops.dissolve_limit(
         bm, angle_limit=math.radians(0.1), verts=bm.verts, edges=bm.edges
     )
+
+    # Clean up degenerate faces with invalid normal vectors
+    zero_normal_faces = [f for f in bm.faces if f.normal.length < 1e-6]
+    if zero_normal_faces:
+        try:
+            bmesh.ops.dissolve_faces(bm, faces=zero_normal_faces)
+        except Exception as e:               
+            pass # Could not dissolve faces
+
+    # Find the most common normal direction, and use that to fix faces that are reversed
+    normals = [f.normal.copy() for f in bm.faces if f.normal.length > 1e-6]
+    if normals:
+        avg_normal = sum(normals, Vector((0,0,0))) / len(normals)
+        wrong_way = [f for f in bm.faces if f.normal.dot(avg_normal) < 0]
+        if wrong_way:
+            bmesh.ops.reverse_faces(bm, faces=wrong_way)
 
     if existing_mesh:
         existing_mesh.clear_geometry()


hlorus:
Looks good, thanks for the PR!

Issue #12, opened 2022-02-13
Snapping to outside geometry


theoryshaw:
**Motivation / Existing Problem**
When creating new geometry, often times you use the context of other objects to snap to. 

**Description**
When in sketch mode, be able to snap to existing geometry outside the sketch

**Alternative Solutions**
When drawing a sketch anew or editing an existing one, be able to snap vertices of the sketch to existing objects outside the sketch. Or an option, anyway, to toggle on/off the geometry outside the sketch. 

**Mockup**
![image](https://user-images.githubusercontent.com/507113/153768896-82a70921-ff99-41c1-9af1-f4f8546f4e4d.png)

Intuos5:
In such cases, there should then also be the option to project the point to which the sketch is snapped to the active sketch.

hlorus:
Implemented: while drawing in a sketch you can snap to geometry outside it. Enable Blender's snapping with the vertex / edge / edge-midpoint / face-midpoint options and it snaps to meshes and other curve objects (Shift bypasses it). Please retest on the latest prerelease build. The follow-up of projecting the snapped point onto the active sketch is tracked with the reference/projection work (#36, #566).

PR #547, opened 2026-05-03
Snapping to 3d geometry

NeilHarbin0:
All of the code in this PR was done by codex AI, and tested in Blender by me.

This feature allows you to snap sketch geometry to existing 3d geometry in the scene. It doesn't cause any constraints to be made, or link anything to make it more parametric unfortunately. It snaps to a projection of 3d geometry onto the workplane. Ideally you would be doing this in an orthographic view normal to the sketch plane, but you don't have to be. If you are not, you can still snap accurately to things that are in the same plane as the workplane, which of course covers the face you made the workplane on.

I believe this covers #12.

This implementation enables:
- Snap to vertexes, edges, edge midpoints (when any of such snap target is enabled, and the "Snap during transforms" blender toggle is on) of visible geometry
- Snap to any if x-ray is enabled, otherwise just front geometry
- Circles can snap tangent to edges, or coincident with vertexes (not constrained though)
- Works with moving points, not just drawing them
- Applies to any drawable thing that makes a point, so lines, start of circle, arcs, points

CAD Sketcher Snapping Feature Demo

Ideally it would show the little icons you get when blender does the snapping itself, but I wasn't able to figure out how to get that without having to implement new rendering just for the plugin which would basically duplicate whatever that existing code is.

### DIFF
--- a/operators/add_arc.py
+++ b/operators/add_arc.py
@@ -8,11 +8,12 @@ from ..declarations import Operators
 from ..stateful_operator.utilities.register import register_stateops_factory
 from ..stateful_operator.state import state_from_args
 from ..solver import solve_system
+from ..utilities.geometry import intersect_line_sphere_2d
 from ..utilities.math import pol2cart
 from .base_2d import Operator2d
 from .constants import types_point_2d
 from .utilities import ignore_hover
-from ..utilities.view import get_pos_2d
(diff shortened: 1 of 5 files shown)


hlorus:
Hey thanks for the PR, this works really well!

PR #548, opened 2026-05-03
Enable deletion of entity and all entities that depend on it

NeilHarbin0:
This code was written by codex AI with some minor assistance from claude, and tested in Blender by me.

The goal was to stop getting the error popup every time I try to delete a point joining two lines. I prefer if it just goes ahead and deletes the lines that depend on the point. Any time I got the previous popup, I have just had to go manually deleting those things myself, which I felt was unnecessary friction. We have undo/redo, so deleting too much generally isn't a concern for me.

Instead of giving the pop-up warning, it gets all the things that depend on it, deletes those too, and gives a message indicating it deleted some extra entities.

CAD Sketcher Entity Delete All v2

### DIFF
--- a/operators/delete_entity.py
+++ b/operators/delete_entity.py
@@ -11,6 +11,7 @@ from ..utilities.ui import show_ui_message_popup
 from ..utilities.data_handling import (
     get_constraint_local_indices,
     get_entity_deps,
+    get_flat_deps,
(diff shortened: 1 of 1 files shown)


hlorus:
Hey thanks for the PR! I agree, with undo/redo we don't really need to be this careful so that change makes sense.

PR #550, opened 2026-05-23
bonsai-integration: add polyline/TAG/GUID, linked sketch, Plan/Elevation workflow, IFC integration

falken10vdl:
This is a long PR covering many aspects to integrate CAD_Sketcher in a BIM IFC Workflow namelyin Blender Bonsia BIM.
The initial goal was to enable practical CAD_Sketcher usage in BIM IFC workflows, particularly with Blender Bonsai BIM.

However, I think that it delivers generally useful CAD_Sketcher improvements (workflow speed, structure, naming, and automation) that also help non-BIM users and other integration targets.

1. BIM-oriented sketch semantics (Plan/Elevation)
Sketch roles are formalized around Plan and Elevation, matching common BIM authoring/documentation intent.
New sketches are auto-tagged based on workplane orientation, reducing manual setup and role mistakes.
Sketch role selection is exposed in the UI through preset search, making BIM intent easier to assign consistently.
The idea of plan/Elvation however is very gneral and can be simply applied to general CAD

2. Linked sketch workflow for multi-view authoring
Added linked sketch creation from source lines, including reference geometry transfer.
Supports projecting linked references from same-role planes, enabling controlled cross-plane drafting.
Maintains geometric relation between source and linked sketches by synchronizing derived workplane behavior.
This is a rather simple "external geometry" support. However it is safe and very powerful

3. Polyline as a first-class entity
Added explicit polyline entity support with stable segment grouping and closed/open state.
Added polyline creation tools and auto-grouping flows in drawing operations.
Polyline (closed or open) is many times usefull in order to speedup workflow so tohre addons can proccess groups opf lines in batch

4. TAG/GUID metadata readiness improvements
These properties are very general and allow any other software to have keys to proccess information from CAD_Sketcher. In the case of IFC this relates to GUID and IFCElemtTypes

5. UX polish
A number of small improvments to make easier to find one the way arround are provided.

Linked Sketch Workplane Behavior (Plan/Elevation)
When creating a linked sketch from a selected source line:

- Local X is aligned with the source line direction.
- Initial local Y uses the source sketch normal.
- Local Z is computed from the cross product of X and Y, then Y is re-orthogonalized from Z and X for numerical stability.
- If the source sketch is tagged Elevation, local Z is flipped (sign inversion), and Y is recomputed from the flipped Z.
- This ensures stable and predictable orientation when transitioning between Plan and Elevation contexts.
- Sketch role is toggled accordingly:
- Plan source creates Elevation linked sketch.
- Elevation source creates Plan linked sketch.
- This same frame computation is also reused during update/sync, so linked sketches remain consistent when source geometry changes.

Linked workplanes also support optional cross-plane reference incorporation: when creating a linked sketch, intersections/projections against other planes with the same sketch role can be added as linked geometry, ordered from closest to farthest relative to the source direction. This allows the new linked sketch to include progressive contextual references from neighboring same-role planes (for example elevation-to-elevation or plan-to-plan), improving alignment and continuity across stacked/parallel modeling planes.

The goal of this workflow is to keep Plan and Elevation sketches consistently oriented for BIM-style authoring.
A Plan sketch generates linked Elevation sketches that grow vertically upward and are oriented as exterior-facing views (seen from outside the model).
Conversely, an Elevation sketch generates linked Plan sketches that are created inward (toward the model), giving a top-view representation tied to that elevation context.
In practice, this means linked sketches are always interpreted as either:

Elevations viewed from outside the model, or
Plans viewed from above.

Cheers!


hlorus:
Hey thanks for the PR, IFC and CAD Sketcher would sure be a great fit! This PR is quite big, are you planning to split out some parts or merge it all together? 
I'd like to avoid cluttering the addon ui with too much stuff that is specific to one workflow. Could we move the "IFC Integration" toggle into the addon preferences and also hide IFC specific stuff if the integration is disabled?

hlorus:
I don't really get the polyline part. So far all entities represent their direct solvespace counterpart, polyline seems to be more of a high-level concept which just groups a set of entities together. Have you considered any alternatives? Maybe this could also be achieved with a generic tagging system like you've done with the sketches.

falken10vdl:
> Hey thanks for the PR, IFC and CAD Sketcher would sure be a great fit! This PR is quite big, are you planning to split out some parts or merge it all together? I'd like to avoid cluttering the addon ui with too much stuff that is specific to one workflow. Could we move the "IFC Integration" toggle into the addon preferences and also hide IFC specific stuff if the integration is disabled?

Indeed! I am playing with the PR and polishing. Yes let me move all IFC related stuff to be toggled just with one simple addon setting Thanks!

falken10vdl:
> I don't really get the polyline part. So far all entities represent their direct solvespace counterpart, polyline seems to be more of a high-level concept which just groups a set of entities together. Have you considered any alternatives? Maybe this could also be achieved with a generic tagging system like you've done with the sketches.

Yes, It is a high level concept that groups entities conveniently as you say. Your summary is spot on. Let me think about this better. Since entities are very much tied to the solver engine, probably as you say I should remove tagging from entities, and have it for both sketches and "high level" groups. Thanks for the feedback! Let me redesing all that part so I really do not touch the entities sub-pannel and create an IFC-entities panel that allows for that (which would always be a subset of all the entities).
Cheers!

falken10vdl:
@hlorus what is the roadmap you guys have in mind for "linking" sketches?
I have implemented something that I think is simple but effective. Just one line in one sketch is the basis for the Xlocal of another linked orthogonal sketch. Doing it orthogonal I think is not a major restriction.
I see two functionalities at least that maybe are general not just tied to IFC:
- linked sketches
- GUID/TAG fileds at sketch level

Then there is this one about high-level grouping of entities which is very convenient in IFC (for example a polyline,that is a succession of lines, tagged IFCWall in a Plan Sketch can be understood as the generator axis for a wall run, while the same, if ti is closed and in an elevation sketch, can be interpreted as a single wall shape) and maybe also convenient for other activities as long as you hav ethe option to create/edit groups of entities and give them TAGs.

Cheers!

hlorus:
> @hlorus what is the roadmap you guys have in mind for "linking" sketches?

We already allow using entities from another sketch when both sketches share the same workplane.

For entities where thats not the case the plan would be to allow the user to create linked and projected entities. So we would probably have linked variants of all (or most) entity types, e.g. SlvsPoint2dLinked or similar. 

Such a system would ideally also allow picking native blender geometry.

PR #553, opened 2026-05-29
Fix DOF stale-status messaging

falken10vdl:
When adding new geometry the DOF info in the UI was missleading. I guess the reason to not  set needs_solve  when creating geomtry is due to performance reasons.

Threfore, I have added a geometry_solved flag to track stale solve state so there is an alert in bpy.ops.view3d.slvs_update() for the user to be warned.
Also removed some 6DOF cap which I guess it is fine for a single 3d rigid body but for the whole sketch system, the DOF can grow much higher.

Fianlly, add_arc has been changed to look exactly the same as the other equivalent files

Cheers!

### DIFF
--- a/model/sketch.py
+++ b/model/sketch.py
@@ -50,7 +50,12 @@ class SlvsSketch(SlvsGenericEntity, PropertyGroup):
     solver_state: EnumProperty(
         name="Solver Status", items=global_data.solver_state_items
     )
-    dof: IntProperty(name="Degrees of Freedom", max=6)
+    dof: IntProperty(name="Degrees of Freedom")
+    geometry_solved: BoolProperty(
+        name="Geometry Solved",
+        description="Whether the sketch geometry has been solved after the last edit",
+        default=True,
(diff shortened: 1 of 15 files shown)


hlorus:
Shouldn't we cover other places which modify geometry to set the 'geometry_solved' flag? Like the tweak operator, delete, trim etc.?

falken10vdl:
@hlorus you are right! Sorry I just realized about it when I was creating arc, lines, circles... 

I have taken a look and have added similar code in Bevel, Move, Delete Entity, Delete Constraint, Tweak, Trim and  Offset (Is any other one missing?). For Tweak, Trim and Offset there is no solve_system so it is similar to what I did in the previous commit. However for the other (Bevel, Move, etc) I read from the code that immediate solve_system is triggered and I guess that makes sense. For those I have just set the flag before so if the solver is not successful we maintain the flag as False.

Please let me know if you think anything else is required.

Thanks!

PR #555, opened 2026-06-03
Dedicated operator for merging points and allow coincident points constraints

falken10vdl:
This is a proposal to allow both to merge points and also coincident constraint them.
Currently adding a coincident constraint to points is merging them which is a destructive process. 

There are sitiuations in which coincident constraining points is required. 
For exmaple in AEC one would like to draw high level structures (like several lines that form a slab, wall run, covering, etc.) that are logically touching each other by means of point coincident constraint. Currently, since this is merging the points, this breaks the independenc of such constructs which is nice should one want to change that constraint in the future for example during a design cycle.

Three commits are provided:

- Let slvs_add_coincident() add constraint for two points (as long as they are not both fixed)
    - Removes the old point-point coincident block in constraints handling and lets coincident be created for point pairs.
    - Adds a guard in operator flow so fixed+fixed point pairs are rejected with a warning instead of creating an invalid constraint.
 

-  Add merge points operator and integrate into UI and keymaps
    - Introduces Merge Points operator with target resolution rules and merge execution.
    - Integrates merge action into declarations, tools UI, and keymaps.
    - Adds merge utility module for pointer rewiring and collapsed-line cleanup.

- Reset geometry solved status when merging points in active sketch just in case solver fails
    - Sets geometry_solved to False before solve. This makes state safer and more consistent if solve is skipped/aborts.
   
Cheers!

### DIFF
--- a/declarations.py
+++ b/declarations.py
@@ -52,6 +52,7 @@ class Operators(str, Enum):
     DeleteConstraint = "view3d.slvs_delete_constraint"
     DeleteEntity = "view3d.slvs_delete_entity"
     InstallPackage = "view3d.slvs_install_package"
+    MergePoints = "view3d.slvs_merge_points"
     Paste = "view3d.slvs_paste"
     Move = "view3d.slvs_move"
     Offset = "view3d.slvs_offset"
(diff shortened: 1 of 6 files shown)


hlorus:
That's a great addition!

Issue #545, opened 2026-04-11
[BUG] Line generated between two points gets constraint to horizontal when it is not.


Koreenium:
### Description

In some scenarios When I want to create a line between two points, it generates a line that is constraint as horizontal but clearly is angled. This also makes some previously made constraints to bug out in the process. Removing horizontal constraint from the line does not fix any issues that have popped up. I have to remove the line all together and make two separate but connected lines and add parallel constraint to them as a work around.

<img width="1128" height="509" alt="Image" src="https://github.com/user-attachments/assets/3c4d9346-4d19-4794-9c9d-d4084782581c" />

Here is also blender file:

[021 B 49501-1 --- Seitenwand_link.zip](https://github.com/user-attachments/files/26648271/021.B.49501-1.---.Seitenwand_link.zip)

### Addon Version

0.27.5

### Blender Version

5.1.0

### What platform are you running on?

Linux

falken10vdl:
@Koreenium Thanks a lot for the feedback! The attached file was of great help and a number of issues where found. 
I have raised PR https://github.com/hlorus/CAD_Sketcher/pull/556 to fix them.
I have introduced a toggle at tool level and also the posibilty to turn off auto constraints if you hold the SHIFT key (regardless of the toggle).

https://github.com/user-attachments/assets/d9669f23-f759-4804-8aa6-2d5a2f7c4c3e

If you want to test it already you can do so by loading CADSketcherPR that will have this loaded. See https://github.com/hlorus/CAD_Sketcher/issues/551

Cheers!

falken10vdl:
@Koreenium The PR to solve all the issues above is not merged into the main branch.
You can directly test it by downloading the latest build as per extension installation instructions:
https://github.com/hlorus/CAD_Sketcher#extension-installation

<img width="873" height="234" alt="Image" src="https://github.com/user-attachments/assets/d15fcc17-9b21-4b67-aa1b-e99a26096cf7" />

that is: https://github.com/hlorus/CAD_Sketcher/archive/refs/heads/main.zip

Please let us know if you see any issue and if not I kindly ask you to get this issue closed :)

Thanks!

Koreenium:
Thank you!

Holding down shift while drawing a line works and also toggle to turn off auto constraints works.

It would be nice if it automatically detected a problem with a constraint that is not plausible and simply didn't add one. But regardless, current solution works. Thank you again!

falken10vdl:
[...]

PR #556, opened 2026-06-04
Fix interactive solving and constraint stability

falken10vdl:
This PR bundles a set of stability improvements around interactive sketch editing and inferred constraints.

This PR was prompted while investigating the behavior shown in issue https://github.com/hlorus/CAD_Sketcher/issues/545:

A  umber of errors where ocurrign apart from teh auto constraint beahviour commented:
 - Deferred solver execution while stateful interactive operators are running, so solving is not triggered mid-edit.
- Fixed recursion in dimensional constraint initialization/settings updates.
- Hardened dimensional constraints against temporarily unresolved entity pointers during reference/initialization transitions.
- Unified inferred auto-constraint behavior:
  - Added a global Auto Constraints toggle for inferred constraints.
  - Added temporary Shift bypass for inferred constraints during interactive placement.
  - Applied this rule where it makes sense for inferred constraints, while keeping structural tool constraints intact (for example rectangle’s defining constraints).

Cheers!

### DIFF
--- a/global_data.py
+++ b/global_data.py
@@ -25,6 +25,7 @@ highlight_entities = []
 
 needs_solve = False
 needs_redraw = False
+stateful_op_running = False
 
 Z_AXIS = Vector((0, 0, 1))
 
--- a/handlers.py
+++ b/handlers.py
@@ -65,6 +65,9 @@ def on_depsgraph_update(scene, depsgraph):
     from . import global_data
 
     if global_data.needs_solve:
+        if global_data.stateful_op_running:
+            return
+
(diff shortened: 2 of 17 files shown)


hlorus:
I really like the auto-constraint toggle! Just one concern: stateful_op_running can get stuck if an operator raises before _end() is reached, silently disabling  the solver for the session. Consider wrapping the invoke/execute paths with try/finally to ensure the flag is always cleared.

falken10vdl:
@hlorus Indeed! Added try/finally guard as suggested. Thanks!

Issue #544, opened 2026-03-25
[BUG] Controllers association to dimensional constraints go wrong when creating/deleting constraints


flboudet:
### Description

I use controllers to provide constants to dimensional constraints.
When I delete a constraint in the middle of the list, the controllers applied further in the list "shift" to the previous constraint.
It looks like the controllers are associated to the index in the list rather than to the constraint itself, leading to a misassociation when the order of the list changes.

https://github.com/user-attachments/assets/a2bacef2-e391-4958-911c-bfc51abe2a68

### Addon Version

0.27.5

### Blender Version

5.1.0

### What platform are you running on?

Linux

hlorus:
Resolved in #557 by @falken10vdl, thanks!

Issue #318, opened 2023-01-11
[BUG] Delete dimensional constraint with drivers


Franckapik:
### Contact Details

fanch44@hotmail.com

### Description

When we delete dimensional constraint with drivers added, the values for all the list of dimensions are modified. 
There is a kind of shift .

![previewDrivers](https://user-images.githubusercontent.com/21478769/211748440-1779c9ef-c0cf-442b-85e0-c9cde43eb5e3.gif)

### Addon Version

0.20

### Blender Version

3.3.1

### What platform are you running on?

Linux

Franckapik:
Hi @hlorus ! 
Thank you for having reported this issue in the good place ! 😉
It seems that we don't have any news? Nobody is working on this at this time, isn't it!?

hlorus:
Yeah it might take a while but i guess you can ask there about the progress...

ecosky:
Just encountered this bug myself lately when working on [some improvements](https://github.com/hlorus/CAD_Sketcher/pull/442) to driver related workflows.

Because that bug has been open for a while and there's no owner, project or milestone for it I think it's unlikely the Blender team will get to this problem soon. I would suggest working around it for now by manually destroying drivers associated with constraints before deleting them, but to my surprise there doesn't appear to be a Blender API that can do this. It seems the existing driver_remove function only removes drivers that match an input name, and there is no API to remove all drivers from what I could tell (see [blender source for driver_remove](https://projects.blender.org/blender/blender/src/commit/b92443e1d9c39ff869bf5662663e089fb7d13772/source/blender/python/intern/bpy_rna_anim.cc#L625)). There's also no way to enumerate the drivers to do get the information necessary to do individual driver_remove calls.

It might be worth submitting a patch to blender to provide a "driver_remove_all" function that would enable scripts to manually remove all drivers as part of their script's other functionality, while at the same time enabling people to work around this old problem in Blender.

hlorus:
Fixed by the native-curves rewrite (dimension values are now stored per-constraint by uid, so deleting one no longer shifts the others). Closing.

PR #557, opened 2026-06-08
Propose index-to-uid migration plan

falken10vdl:
@hlorus 
I was looking to https://github.com/hlorus/CAD_Sketcher/issues/544 and found that a way to solve it is to remap indexes when delete is being done, however there might be other situations in which the items in the list are reordered and remapping would be required.
Please take a look to the attached file with a proposal to move from index to UID for both entities (which I think will suffer a similar problem although smaller as there is slvs_index) and constraints to make sure referencing them is stable regardless of the ordering.
Is this something worth doing? 
I am proposing here without showing any code yet as it looks to me this will have to touch quite a bit of code and woudl like to knwo if this is somthign that wants to be impolemented at all first :)

Cheers!

### DIFF
--- a/ui/panels/constraints_list.py
+++ b/ui/panels/constraints_list.py
@@ -41,7 +41,7 @@ def draw_constraint_listitem(
             uid = getattr(constraint, "constraint_uid", "")
             key = None
             if uid:
-                key = context.scene.sketcher.get_or_create_constraint_value_endpoint(
+                key = context.scene.sketcher.get_constraint_value_endpoint(
                     constraint
                 )
             if key:
--- a/ui/panels/constraints_list.py
+++ b/ui/panels/constraints_list.py
@@ -37,6 +37,16 @@ def draw_constraint_listitem(
     middle_sub = row.row()
 
     for constraint_prop in constraint.props:
+        if constraint_prop == "value":
+            uid = getattr(constraint, "constraint_uid", "")
+            key = None
+            if uid:
+                key = context.scene.sketcher.get_or_create_constraint_value_endpoint(
+                    constraint
+                )
+            if key:
+                middle_sub.prop(context.scene, f'["{key}"]', text="")
+                continue
         middle_sub.prop(constraint, constraint_prop, text="")
 
     # Context menu, shows constraint name
--- a/__init__.py
+++ b/__init__.py
@@ -8,7 +8,7 @@ from bpy.app import background
 bl_info = {
     "name": "CAD Sketcher",
     "author": "hlorus",
-    "version": (0, 27, 6),
+    "version": (0, 27, 7),
     "blender": (4, 0, 0),
     "location": "View3D > Toolbar",
     "description": "Parametric, constraint-based geometry sketcher",
--- a/blender_manifest.toml
+++ b/blender_manifest.toml
@@ -1,7 +1,7 @@
 schema_version = "1.0.0"
 
 id = "CAD_Sketcher"
-version = "0.27.5"
+version = "0.27.7"
 name = "CAD Sketcher"
 tagline = "Parametric, constraint-based geometry sketcher"
 maintainer = "hlorus <dave19924@gmail.com>"
--- a/experiments/test_driver_stability_uid.py
+++ b/experiments/test_driver_stability_uid.py
@@ -1,0 +1,82 @@
+"""Test driver stability using UIDs (fix for issue #544).
+
+With UID-based driver paths like scene["slvs:c:{uid}"], constraint values
+are accessed by their unique identifier rather than collection index.
+This ensures drivers remain stable even when other constraints are deleted.
+"""
+import bpy
+
+scene = bpy.data.scenes.new("DriverTestUID")
+bpy.context.window.scene = scene
+
+sse = scene.sketcher.entities
+ssc = scene.sketcher.constraints
+
+sse.ensure_origin_elements(bpy.context)
+wp = sse.origin_plane_XY
+sketch = sse.add_sketch(wp)
+scene.sketcher.active_sketch_i = sketch.slvs_index
+
+# Create 3 distance constraints
+p0 = sse.add_point_2d((0, 0), sketch, fixed=True)
+
+p1 = sse.add_point_2d((1, 0), sketch)
+c1 = ssc.add_distance(p0, p1, sketch=sketch, init=True)
+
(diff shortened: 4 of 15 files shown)


hlorus:
I'm afraid this wont change anything since slvs_index is already persistent. This is a bug on blender's side which has been reported quite a while back, see: https://github.com/hlorus/CAD_Sketcher/issues/318

falken10vdl:
@hlorus wow! I am just new to drivers (did not know about them until this morning :) ) and this error is dragging since 4 years ago... 
I Guess we can do a workarround remapping datapath/indexes for the time being since this might take long to get solved? What do you think?

hlorus:
How would you do that? I'm not that familiar with the driver API. But sure if you find a workaround that would be nice!

falken10vdl:
@hlorus I dug a bit deeper and this looks very similar to https://blender.stackexchange.com/questions/346658/group-node-input-order-change-affects-data-path-and-breaks-drivers
The key point from that thread is to use stable names/ids in driver paths instead of positional indices (answer 1, but answer 2 is a little bit of contrary, but dependent on Blender to fix it...).

For CAD Sketcher, I think the safest direction is still stable string uid data paths.

- Constraints: scene.sketcher.constraint_values[uid].value style addressing is the robust approach. Current constraint deletion is plain collection removal, so index-based paths shift and Blender does not remap those driver paths for us.

- Entities: entities (slv_index) are better than constraints, but not fully safe for drivers. Entity removal does swap-with-last plus pointer remap, so internal CAD Sketcher references are repaired, but Blender driver strings are outside that repair pass, so drivers targeting the moved last entity can still become wrong after delete. So yes, this limits impact mostly to the deleted slot + previous last slot (better than full reindex shifting), but it is still not fully safe for all driver-attached identities.

What do you think? Maybe start addressing the constraints case by using constraint_uid so we get https://github.com/hlorus/CAD_Sketcher/issues/544 already solved?

falken10vdl:
> How would you do that? I'm not that familiar with the driver API. But sure if you find a workaround that would be nice!

Well in constraints_list.py you have:
```
    for constraint_prop in constraint.props:
        middle_sub.prop(constraint, constraint_prop, text="")
```

For dimensional constraints, constraint_prop is value (defined as `props = ("value",)`):

Because the owner is the actual constraint item inside the collection, Blender resolves driver path from that item’s RNA location in the collection, so you get paths like:
bpy.data.scenes["Scene"].sketcher.constraints.distance[1].value

The idea would be to stop binding the UI value widget directly to the constraint collection item, and instead bind it to a stable UID endpoint:

1. Add constraint_uid on each constraint (persistent, unique).
2. Add a scene-level UID-keyed endpoint collection, for example: scene.sketcher.constraint_values["c_xxxxx"].value
4. In constraints_list, resolve endpoint by constraint_uid and draw endpoint.value in the row (instead of constraint.value).
5. Keep endpoint.value <-> real constraint.value synchronized in update callbacks.
6. Move operator/gizmo resolution from type+index to constraint_uid where possible.

This way  when user adds a driver from the constraints UI, the path targets the UID endpoint  scene.sketcher.constraint_values["c_xxxxx"].value, not constraints.distance[index].value (fragile on reorder/delete).

Cheers!

hlorus:
Thanks for the digging! That sounds like a good approach for a workaround!

falken10vdl:
@hlorus Please take a look to the PR:

-   UIDs to all dimensional constraints via constraint_uid property (I just use a  16-character string like d254205dab6c3315 not a formal RFC-4122 UUID string which is much longer but let me know what you think is best) 
- Scene-level ID properties at bpy.data.scenes["Scene"]["slvs:c:<uid>"] (exmple: bpy.data.scenes["Scene"]["slvs:c:d254205dab6c3315"]) as driver targets instead of using constraint RNA paths that are fragile if using indexes
- UI binding in property panels and constraint drawer to use the scene endpoint
- I bumped to version v0.27.7 that auto-assigns UIDs and creates endpoints for existing files
- Syncs bidirectionally: constraint value changes update the endpoint via constraint_value_update_cb, and driver-written endpoints are pulled back into constraints in the depsgraph handler

Cheers!

ui/panels/constraints_list.py
```diff
@@ -37,6 +37,16 @@ def draw_constraint_listitem(
     middle_sub = row.row()
 
     for constraint_prop in constraint.props:
+        if constraint_prop == "value":
+            uid = getattr(constraint, "constraint_uid", "")
+            key = None
+            if uid:
+                key = context.scene.sketcher.get_or_create_constraint_value_endpoint(
```
hlorus: We should not call this from a draw function, writing to ID during draw is a blender anti-pattern. The key should be created on constraint init and only read here.

hlorus (changes requested):
The main thing here is that with this change there are two sources for constraint values that have to be kept in sync. It would be great if we could still just have a single source of truth. Maybe such that constraint values are just always stored in the new value collection and value constraints only store a pointer to where the value is. That should simplify the changes in this PR a bit because the sync logic is no longer needed.

falken10vdl:
@hlorus Agree. I have removed the `value_store` (only kept for the migration process) and now `scene["slvs:c:<uid>"]`  is written at constraint creation `_get_value` reads directly from `scene[key]` — no `value_store` involved
Also, sorry I keep forgetting not to write in drawing modes. That is also fixed.

Note that I have just done this in scene properties. I guess another option could be adding a per-sketch Empty, which would allow to move the ID property from the Scene to the sketch (Empty) object. What do you think would be best?

hlorus:
Sorry for the long wait. This seems to work well but some of the tests fail: 
 - test_distance — line length 0.0 instead of 2.0
 - test_diameter — radius 1.5 instead of 3.0
 - test_angle — point not moved
 - test_distance_flip — point not flipped

The cause seems to be: _init_constraint creates the scene property with init_value=0.0 (from unset value_store). Then _get_init_value in distance/diameter/angle finds the scene key already exists and returns 0.0 instead of computing from entity geometry.


Edit: I've added a test case to cover the reported issue in #544

falken10vdl:
@hlorus Thanks for the feedback. Sorry I forgot to run the tests. Added some fixes in angle, distance and diameter to address the errors. I have also added a test in experiments similar to the one you had created but using UIDs that are stable when removing the driver.
Cheers!

hlorus:
Great, thanks!

PR #559, opened 2026-07-02
Native Curves Refactor: Blender Curves as data source

hlorus:
This is another attempt improving the addon's data model, following #522 and a few other experiments.
This approach is possible thanks to some relativley new changes in blender, mainly geometry nodes and attributes for curve objects. 
> [!IMPORTANT]
> Due to the dependency on new blender features the minimum supported blender version is: 4.3
 
Until now, the addon has maintained its own data storage. That approach led to a lot of duplicated logic and not ideal integration with Blender. Moving to Blender's native data changes that, with big improvements to performance and stability, and it potentially clears the way for some features while at the same time simplifying the addon alot.

Since this changes the fundamental data storage alot of places are effected which results in this massive PR.

## Summary

Replaces the custom entity PropertyGroup storage system with native Blender Curves as the source of truth for sketch geometry. Sketches are now Curves objects, constraints are stored per-sketch.

### Architecture

```
Workplane (Empty, hidden)
  └── Sketch (Curves Object)
        ├── Curve attributes: curve_id, sketch_type, relationships, flags
        ├── sketch_constraints: CollectionProperty (per-sketch)
        └── Custom props: is_cad_sketch, solver_state, dof
```

### Key Changes

**CurveRef API** (`model/curve_ref.py`)
- Typed subclasses: `PointRef`, `LineRef`, `ArcRef`, `CircleRef`
- Read/write/create/delete via clean API
- `PointRef.co` setter auto-rebuilds referencing segments
- `batch_update()` context manager for bulk operations

**Sketch = Curves Object** (`model/sketch_ref.py`)
- No intermediary PropertyGroup — the Curves object IS the sketch
- `Sketch(obj)` accessor for `.constraints`, `.workplane_object`, `.solver_state`
- Discovery via `get_sketches()`, activation via `get_active_sketch()`

**Per-Sketch Constraints** (`bpy.types.Curves.sketch_constraints`)
- Constraints stored on the Curves data block, not globally
- Delete sketch = delete constraints automatically
- No orphan tracking needed

**CurveSolver** (`curve_solver.py`)
- Reads geometry from curve attributes, writes back via `rebuild_segments()`
- Replaces legacy entity-based solver entirely

**Drawing**
- Overlay reads positions from point curves via relationship attributes
- No stale segment copies — lines/arcs computed from source points each frame

### What Works
- Create/edit sketches on all three workplanes
- Draw points, lines, arcs, circles, rectangles
- All 13 constraint types (coincident, distance, angle, diameter, etc.)
- Constraint solving with live feedback
- Tweaking (drag), moving (G key), copy/paste, select all/invert/extend
- Delete geometry and sketches
- Context menu with flag toggles
- Continuous line drawing with auto-constraints

### ToDo

- [x] Bevel/offset/trim operators (still entity-based)
- [x] 3D operators (add_point_3d, add_line_3d, add_workplane)
- [x] File versioning migration (old → new format)
- [ ] Cleanup, legacy entity classes kept registered for file compatibility
- [ ] Snapping
- [x] Handle manual/native edits to sketch curves
- [ ] Various bugs to fix
- [ ] Polish

> [!WARNING]
> This is a draft, do not save important files


hlorus:
@falken10vdl any input on this? There might be downsides to this approach when integrating CAD Sketcher within other addons..

falken10vdl:
@hlorus I think that this PR is the right architectural approach. For sure the current bonsai integration will need to be refactored mostly from scratch but I think it will then be cleaner and simpler.

What I think is that maybe in the PR only having curve_id will not be completely safe in some cases.

 I would propose:
- Implement UUID as source of truth (similar to the constraint PR) to keep UUID as the authoritative persistent identity for integration, serialization, and cross-addon mapping.
- curve_id as operational index. Use curve_id as a fast local handle for solver and geometry ops inside one sketch/session.

Here I think having both IDs it is better idea than only one (like in the constraints). Not two sources of truth but rather one  truth (UUID) plus one accelerator.

Cheers!

hlorus:
Thanks for the input!
Indeed a UUID is a good idea, probably also required for future cross-sketch geometry referencing. I'll have a look into that.
Cheers

hlorus:
@falken10vdl i did switch to UUIDs and it works well. Only issue i've found is with string attributes, they don't seem to survive some operations of native geometry tools. Which can lead to a corrupted state. With multiple INT attributes we can workaround this issue.

falken10vdl:
@hlorus Nice development! Do you have an example of geometry tool that would corrupt the uuid stored as string attribute?
I guess in any case a  int32_2d which gives you 64bits (or two of them) could maybe be enough.
Thanks!

hlorus:
> @hlorus Nice development! Do you have an example of geometry tool that would corrupt the uuid stored as string attribute? 

Yeah this script demonstrates the issue when pasting and running in the blender text editor:
```
import bpy

# A Curves object with a STRING id and an INT id on 3 single-point curves.
cv = bpy.data.hair_curves.new("demo")
cv.add_curves([1, 1, 1])
cv.attributes.new("str_id", "STRING", "CURVE")
cv.attributes.new("int_id", "INT", "CURVE")
for i in range(3):
    cv.attributes["str_id"].data[i].value = b"id-%d" % i
    cv.attributes["int_id"].data[i].value = i

obj = bpy.data.objects.new("demo", cv)
bpy.context.scene.collection.objects.link(obj)

def show(tag):
    s = cv.attributes.get("str_id")
    print(tag, "str_id:", "MISSING" if not s else [d.value for d in s.data],
          " int_id:", [d.value for d in cv.attributes["int_id"].data])

show("before:")

# Delete one point with the native Edit-Mode Delete tool.
bpy.context.view_layer.objects.active = obj
sel = cv.attributes.get(".selection") or cv.attributes.new(".selection", "BOOLEAN", "POINT")
sel.data[0].value = True
bpy.ops.object.mode_set(mode="EDIT")
bpy.ops.curves.delete()
bpy.ops.object.mode_set(mode="OBJECT")

show("after: ")
```

> I guess in any case a int32_2d which gives you 64bits (or two of them) could maybe be enough. Thanks!

Fair input, i'll have a look.

hlorus:
@falken10vdl i've switched to two int32_2d as you've suggested.

falken10vdl:
@hlorus Thanks for the explanation! I have raised a bug to blender based on the input above https://projects.blender.org/blender/blender/issues/161836
Thanks!

falken10vdl:
WoW! Can't wait to use it! 
Thanks!

phkahler:
@hlorus  Are you going to fully remove dependence on the solvespace constraint solver? That would make me a little bit sad ;-)  But glad to see these improvements and Blender getting new goodness!

hlorus:
@phkahler no not at all, solvespace is still used to solve the constrained system. The PR just refactors the data model of the addon. Entities are now represented as native blender curve geometry, constraints are still stored the same as before (CollectionProperty). Anyway when solving the system that data is converted and sent to the solver module.

PR #602, opened 2026-08-11
Stabilize circle and arc tangent branch selection

franklincg:
## Summary

- Stabilizes circle/arc tangent branch selection.
- Reduces large jumps for unequal-radius curves.
- Adds regression tests for external and internal tangency.

## Validation

- Python compilation passed.
- Only 2 files were changed.
- No conflicts with the main branch.

Fixes #291

PR #601, opened 2026-08-11
Add direct value entry for dimensional constraints

DariusJDebugs:
## Summary

Adds immediate numeric value entry when creating dimensional constraints.

Dimensional constraint operators now expose an optional final `Value` state, allowing users to type the desired value directly during constraint creation while preserving the measured value when no new value is entered.

Applies to:

- Distance
- Diameter
- Angle

## Testing

- Ruff lint: passed
- Ruff formatting: passed
- `git diff --check`: passed
- CAD Sketcher Blender test suite: **206 tests passed**
  - 4 skipped
  - 2 expected failures

Closes #259

PR #603, opened 2026-08-11
Preserve circular tangent branch while dragging

### LINKED ISSUE #291: Tweaks make arcs/circles less jittery with tangent constraint on the inside edge.
When making 2 arc/circles entities tangent, they really want to "face each other".

image

When constraining the circles, it is easier to make them tangent without too much jumping around.

[circle_tangents.blend.zip](https://github.com/hlorus/CAD_Sketcher/files/9487168/circle_tangents.blend.zip)

image

This could be an issue of the solver though.

### Workaround

Considering C1 and C2 the centerpoints of the 2 circles, You can create a point P on one of the circles, then make the point coincident to both circles. Finaly, create a construction line [C1 , P] and make C2 coincident to that line.

image

It is a considerably better experience moving around, but only allows "inner" tangent

DariusJDebugs:
Fixes #291.

Circular tangent constraints could change solution branch when the sketch was
re-solved during interactive dragging. Using the original issue reproduction,
the baseline changes from the internal tangent branch to the external branch
on the first drag step, causing the visible geometry jump.

This change initializes the auxiliary tangent point on the branch closest to
the current geometry, preserving the existing internal/external tangency
solution across solver rebuilds.

A regression test based on the original issue reproduction has been added.

Validation:

- Baseline: regression fails at drag step 1 with a tangent branch change.
- Patched: regression passes with no branch changes.
- Full test suite: 206 tests pass
  (4 skipped, 2 expected failures).

Bounty: #291 / Bounty - Small


hlorus:
Thanks for this @DariusJDebugs — seeding the auxiliary tangent point on the branch closest to the current geometry is a sound approach, and the regression test built from the original repro is exactly right. 🙏

Unfortunately this overlaps with #602, which fixes the same issue (#291) and was merged first, so I'm going to close this one as a duplicate. That's purely a timing call — no reflection on the quality here, which is excellent (and thanks again for #601!).

Really appreciate your continued contributions; hope you'll pick up another one soon.

PR #605, opened 2026-08-13
Support distance constraints on 3D entities

### LINKED ISSUE #55: Support distance constraint in 3d

franklincg:
Fixes #55.

Restores support for scene-level 3D distance constraints alongside native 2D curve sketches and makes SlvsDistance render correctly in world space for 3D entities.

Highlights:
- support 3D points, lines and workplanes;
- compute the correct world-space matrix_basis() for 3D dimensions;
- preserve the existing native 2D curve path;
- restore solving outside an active sketch;
- add regression coverage for 3D distance placement, orientation and solving.

Bounty: Small ($100), claimed in #55 before submission.

### DIFF
--- a/gizmos/base.py
+++ b/gizmos/base.py
@@ -6,12 +6,12 @@ from .utilities import get_color, get_constraint_color_type, set_gizmo_colors
 
 class ConstraintGizmo:
     def _get_constraint(self, context):
-        from ..model.sketch_ref import get_active_sketch
+        from ..model.sketch_ref import get_active_constraints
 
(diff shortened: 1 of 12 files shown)


hlorus:
Closing this — see my note on #55 for the full reasoning. Short version: it builds on the legacy `scene.sketcher.entities` system that the native-curves refactor (#559) is retiring, so we don't want to re-entrench that path. 3D will instead land as a native 3D-sketch concept (Curves as source of truth), tracked in #607. Thanks again for the careful work here, and hope you'll take a look at #607.

PR #606, opened 2026-08-13
Add live project geometry from external objects

### LINKED ISSUE #36: Project lines form objects [REQUEST]

franklincg:
Fixes #36.

Adds a Project Geometry workflow for bringing external object geometry into a sketch while keeping a live reference to the source object.

Highlights:
- project mesh geometry into the active sketch;
- keep projected geometry linked to its source object;
- refresh projected entities when the source mesh changes;
- expose Project Geometry in the tools panel;
- keep projection as a separate operation instead of duplicating source geometry;
- add regression coverage for projection and source synchronization.

This implementation is based on current main and incorporates the maintainer feedback around keeping projected entities linked to their source geometry.

Bounty: Medium ($250), claimed in #36 before submission.

### DIFF
--- a/.github/workflows/.money-agent-trigger-placeholder
+++ b/.github/workflows/.money-agent-trigger-placeholder
@@ -1,0 +1,1 @@
+trigger
--- a/.github/workflows/.money-agent-trigger-placeholder
+++ b/.github/workflows/.money-agent-trigger-placeholder
@@ -1,1 +1,0 @@
-trigger
--- a/declarations.py
+++ b/declarations.py
@@ -69,6 +69,7 @@ class Operators(str, Enum):
     NodeExtrude = "view3d.slvs_node_extrude"
     NodeArrayLinear = "view3d.slvs_node_array_linear"
     NodeRevolve = "view3d.slvs_node_revolve"
+    ProjectGeometry = "view3d.slvs_project_geometry"
     RegisterDrawCB = "view3d.slvs_register_draw_cb"
     RenameCurve = "view3d.slvs_rename_curve"
     Restore = "view3d.slvs_restore"
--- a/declarations.py
+++ b/declarations.py
@@ -1,7 +1,6 @@
 # Please keep this file in alphabetical order
 from enum import Enum
 
-
 # Blender's built-in Select Box tool — the standard tool we return to when
 # leaving one of our tools (ESC / cancel).
 BLENDER_SELECT_TOOL = "builtin.select_box"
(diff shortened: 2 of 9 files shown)


hlorus:
Thanks @franklincg — this is the right direction: projected geometry as real native curves with a live source link. Nice work, and the attribute-backed vertex identity + self-heal are well thought out.

One change I would like before merging: move the binding off the sketch object's custom properties and into the Curves data where it belongs. The refactor's guiding rule is that the Curves object is the source of truth, and three of the four binding fields are plain PODs that fit curve attributes cleanly:

- `vertex_id` -> CURVE-domain `INT`
- fallback `vertex_index` -> CURVE-domain `INT`
- `last_co` -> CURVE-domain `FLOAT_VECTOR`

Stored as curve attributes they auto-reindex and auto-delete with the curve (INT/FLOAT survive `remove_curves`), which removes the orphaned-prop problem entirely — the `stale` / `clear_projected_point_binding` bookkeeping goes away.

The only field that genuinely cannot be an attribute is the source **object** reference (attributes can hold no ID pointer, and a name STRING attr would be dropped by `remove_curves`). Keep that one as a `PointerProperty` — e.g. a small `CollectionProperty` of object pointers on the sketch object with a `project_src_slot` INT curve attribute indexing into it, so points sharing a source share one entry.

Net: `4 x N` stringly-keyed props on the object -> per-curve attributes + one small pointer collection. Same behaviour, fully native, less cleanup code.

franklincg:
Thanks for the review. I’ve implemented the requested binding cleanup:

moved the projection binding POD data to CURVE-domain attributes
kept the source object reference in a small PointerProperty collection
removed the old per-point custom-property bookkeeping
expanded the regression coverage

Current CI is green for lint, perf, Blender 5.0, and Blender latest.

The PR is ready for re-review/merge when you have a chance.

hlorus:
Merged — thanks @franklincg! 🙌 The reworked native-Curves storage (binding metadata as CURVE-domain attributes + the compact source-pointer slot) is exactly the right shape, and the live source refresh works well. Really solid contribution.

This closes a bounty issue (#36, *Bounty – Medium*), so you're welcome to collect it: submit an invoice on our Open Collective at https://opencollective.com/cadsketcher/expenses/new — choose **Invoice**, and reference `Add live project geometry from external objects #606`. Thanks again!

franklincg:
Thanks! Invoice submitted on Open Collective as expense #342262. Appreciate the review and merge! 🙌

PR #609, opened 2026-08-13
Add native CAD Sketcher pie menu (#153)

### LINKED ISSUE #153: Pie Menu
Extensive Pie Menu that can replace completely the use of T-Panel and N-Panel.

Below is a layout with keymaps that have worked well in my testing. Feels speedy and intuitive.

image

Above was created using Pie Menu Editor. Experiencing trouble obtaining the behavior expected for activate sketch / leave sketch portion of pie menu. This particular portion of the pie should invoke solvespace select and active sketch. Upon "leave sketch" it will revert to select tool outside of solvespace and deactivate sketch. Leave sketch works simply given index is always -1.

Currently, sketches can cycle from 0 onward easily. Renaming the sketch works through the pie menu as well.
Some difficulties exist when selecting convert type.

I can provide what I have for a jumping off point should you be interested to add a pie menu. Or at least to test functionality of this pie menu design.

franklincg:
Implements the native built-in pie menu requested in #153 without depending on Pie Menu Editor.

Included:
- native Blender pie menu for core drawing tools
- constraints pie plus dimensional/geometric submenus
- context-sensitive sketch entry/exit using the existing AddSketch / SetActiveSketch flow
- sketch switching, selection, trim, bevel, offset, construction, align-view, and node-tool access
- addon keymap for opening the pie menu
- regression coverage for menu identity/default shortcut

The implementation reuses the current native-Curves operators and sketch-mode transitions, so leaving a sketch returns to Blender select and entering a sketch uses the CAD Sketcher select tool through the existing activation path.

Bounty: Small ($100)

Fixes #153.

### DIFF
--- a/registration.py
+++ b/registration.py
@@ -12,6 +12,7 @@ core_modules = [
     "keymaps",
     "workspacetools",
     "ui",
+    "ui.pie_menu",
     "draw_handler",
     # Register this as last module
     "registration_delayed",
--- a/testing/test_pie_menu.py
+++ b/testing/test_pie_menu.py
@@ -1,0 +1,9 @@
+from unittest import TestCase
+
+from ..ui.pie_menu import PIE_MENU_ID, PIE_SHORTCUT, VIEW3D_MT_slvs_pie
+
+
+class TestPieMenu(TestCase):
+    def test_identifier_and_shortcut(self):
+        self.assertEqual(VIEW3D_MT_slvs_pie.bl_idname, PIE_MENU_ID)
+        self.assertEqual(PIE_SHORTCUT, ("M", True, True))
--- a/ui/pie_menu.py
+++ b/ui/pie_menu.py
@@ -22,6 +22,12 @@ def _tool_set(layout, text, tool_name, icon="NONE"):
     return props
 
 
+def _call_pie(layout, text, menu_id):
+    props = layout.operator("wm.call_menu_pie", text=text)
+    props.name = menu_id
+    return props
+
+
 def _sketch_slot(layout, context):
     sketch = get_active_sketch(context)
     if sketch:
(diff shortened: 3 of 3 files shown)


hlorus:
Thanks @franklincg — nice, self-contained pie and the disabled-state gating on the drawing slots is a good touch. Verified all the operator/tool/property references resolve, so it's solid there.

One UX issue to fix before merge: the constraints sub-pie renders as a **long horizontal row of buttons**, not a radial pie. `VIEW3D_MT_slvs_constraints_pie.draw` calls `layout.menu_pie()`, but it's opened via `slot.menu(...)`, which is a regular submenu — `menu_pie()` only lays out radially when invoked through `wm.call_menu_pie`, so here it falls back to a flat row.

Two acceptable fixes, pick whichever feels better:
- **Keep it a pie:** open it via `wm.call_menu_pie` (e.g. `op = slot.operator("wm.call_menu_pie", ...); op.name = VIEW3D_MT_slvs_constraints_pie.bl_idname`) so it renders radially, or
- **Make it a plain menu:** drop `menu_pie()` and lay the constraints out as a normal `Menu` — that pops up compactly next to the cursor.

Either is fine; the current in-between (pie layout shown as a submenu row) is the thing to avoid.

franklincg:
Thanks for the review — fixed in commit 41df22776d63dd09a2b5008a8426c638c0a5ccab.

I kept the constraints submenu as a radial pie and changed the invocation to use `wm.call_menu_pie`, so `VIEW3D_MT_slvs_constraints_pie` now renders radially instead of falling back to the horizontal submenu row.

The PR remains ready for review. Thanks!

hlorus:
Merged — thanks @franklincg! Clean, self-contained pie menu, and the constraints sub-pie now opens radially as intended. 🙌

This closes a bounty issue (#153, *Bounty – Small*), so you're welcome to collect it: submit an invoice on our Open Collective at https://opencollective.com/cadsketcher/expenses/new — choose **Invoice**, and reference `Add native CAD Sketcher pie menu #609`. Thanks again for contributing!

PR #610, opened 2026-08-14
Fix #67: Improve auto-creating constraints

### LINKED ISSUE #67: Improve auto-creating constraints
Tools that automatically add constraints should check if the system still solves. If there's an error the constraints shouldn't be added.

Additionally it would be great to have the option to turn off automatic constraining with a global toggle and ideally also a modifier key.

DariusJDebugs:
## Summary

Resolves the implementation requested in #67.

**Diagnosed cause:** Issue terms concentrate in: docs/content/code_docs.md, operators/base_stateful.py, curve_solver.py, model/base_constraint.py, docs/content/tools.md. Baseline test command passed, so the reported failure is likely narrower than the full suite or needs specific reproduction state.

## Scope

- `operators/add_line_2d.py`
- `operators/add_rectangle.py`
- `operators/base_stateful.py`
- `testing/test_auto_constraints.py`

## Acceptance criteria

- No explicit acceptance checklist was extracted from the issue.

## Verification

- `blender --background --command extension validate` — **PASS**
- `blender --background --command extension build --output-filepath ./.optix_blender_extension.zip` — **PASS**
- `blender --background --command extension install-file --repo user_default --enable ./.optix_blender_extension.zip` — **PASS**
- `blender --background --python ./scripts/ci_run_tests.py` — **PASS**

## OPTIX review gate

- Phase 2 status: `PHASE2_COMPLETE`
- Critic: `PASS` / `PASS`
- Repair loops: `2`

Closes #67

### DIFF
--- a/operators/add_line_2d.py
+++ b/operators/add_line_2d.py
@@ -64,11 +64,17 @@ class View3D_OT_slvs_add_line2d(Operator, Operator2d):
 
             threshold = 0.1
             if angle < threshold or angle > HALF_TURN - threshold:
-                sketch.constraints.add_horizontal(curve_id_1=line_cid)
-                self.has_alignment = True
+                self.has_alignment = bool(
+                    self.add_auto_constraint(
+                        context, sketch.constraints.add_horizontal, curve_id_1=line_cid
+                    )
+                )
             elif (QUARTER_TURN - threshold) < angle < (QUARTER_TURN + threshold):
-                sketch.constraints.add_vertical(curve_id_1=line_cid)
-                self.has_alignment = True
+                self.has_alignment = bool(
+                    self.add_auto_constraint(
+                        context, sketch.constraints.add_vertical, curve_id_1=line_cid
+                    )
+                )
 
         ignore_hover(line_cid)
         return True
--- a/operators/add_rectangle.py
+++ b/operators/add_rectangle.py
@@ -70,14 +70,23 @@ class View3D_OT_slvs_add_rectangle(Operator, Operator2d):
         return True
 
     def fini(self, context: Context, succeede: bool):
-        if hasattr(self, "lines") and self.lines:
+        if succeede and hasattr(self, "lines") and self.lines:
             sc = self.sketch.constraints
             # Auto axis-alignment constraints (inferred) respect the toggle and
             # Shift bypass; the numeric distance constraints below are explicit.
             if self.use_auto_constraints(context):
                 for i, line_ref in enumerate(self.lines):
-                    func = sc.add_horizontal if (i % 2) == 0 else sc.add_vertical
-                    func(curve_id_1=line_ref.curve_id)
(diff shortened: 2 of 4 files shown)


hlorus:
Thanks @DariusJDebugs. The core idea is right and matches #67. The issue is how the rollback is done, it needs reworking before this can go in.

**The rollback does not follow the stateful-operator paradigm, and it drags existing geometry while drawing.** The framework is snapshot-and-replay: `create_snapshot` at invoke, and every interactive update `_apply_undo` restores that snapshot and re-runs `main()` (see `stateful_operator/logic.py`). `add_auto_constraint` instead hand-rolls its own undo (`add`, `solve`, then `constraints.remove` plus a second `solve`), duplicating machinery that already exists in `base_stateful.py` (`_snapshot_all_curves` / `_restore_all_curves`). Worse, the trial `solve_system` runs inside `main()` on the per-frame path, and `solve_system` is not a pure check, on success it writes solved positions back (`_solve_once` -> `_write_results`). Since the coincident auto-constraint is added from `create_element` every frame, each frame a solve runs and drags under-constrained geometry to follow the cursor.

The feasibility check must be side-effect-free and reuse the framework snapshot, roughly:

```python
def add_auto_constraint(self, context, add, state_data=None, **kwargs):
    if not self.use_auto_constraints(context, state_data):
        return None
    snapshot = self._snapshot_all_curves(context)
    add(**kwargs)
    solvable = solve_system(context, sketch=self.sketch)
    self._restore_all_curves(context, snapshot)   # always restore: no persisted geometry
    if not solvable:
        return None
    return add(**kwargs)   # re-add on clean state; do NOT solve here
```

Restoring in all cases removes the drag on both accept and reject, drops the hand-rolled `remove()`, and lets the operator normal end-of-op solve produce the real result. (A cheaper equivalent is saving/restoring just the point positions via `foreach_get`/`foreach_set` instead of a full snapshot per frame. Or move validation to `fini` entirely, where coincident was already solved before this PR.)

Two smaller points:
- **Coverage: `bevel.py`** (around lines 219-230) also auto-adds `tangent`/`equal` constraints and should go through the same validation. (`add_geometric_constraints.py` is explicit user action, correctly out of scope.)
- Please drop the invented comment in `add_rectangle.py` about "Blender's RNA wrapper does not guarantee stable identity for bound methods", `self.sketch.constraints` is a plain Python accessor.

hlorus:
Thanks @DariusJDebugs. I merged your work, with some changes, in #645 (brought up to date on current main and refined a bit). Your original commits are preserved with your authorship.

This closes a bounty issue (#67, Bounty - Small), so you are welcome to collect it: submit an invoice on our Open Collective at https://opencollective.com/cadsketcher/expenses/new, choose Invoice, and reference `Improve auto-creating constraints #610`. Thanks again for contributing!

PR #611, opened 2026-08-14
Keep generated mesh element identifiers stable through edits (#6)

### LINKED ISSUE #6: Keep mesh vertex/face identificators consistent through edits[REQUEST]
**Motivation / Existing Problem**
The generated mesh gets deleted and recreated on each edit, meaning the geometry can't be reliably linked.

**Description**
The generated mesh should keep consistent identificators.

**Please add possible uses for this feature**

franklincg:
Fixes #6.

Adds deterministic generated-mesh element identity to the current native-Curves conversion path.

- stores `cad_sketcher_vertex_id` on generated POINT geometry;
- stores `cad_sketcher_face_id` on generated FACE geometry;
- identity is regenerated deterministically from element index on every Geometry Nodes evaluation, preserving correspondence for coordinate-only edits while topology is unchanged;
- bumps the programmatic convert-node-group version so existing groups rebuild in place;
- adds regression coverage for generated ID nodes, domains, wiring, zero-ID weld protection, and stale-group upgrade behavior;
- CI passes lint, Blender tests, and perf.

Bounty: Small ($100).

### DIFF
--- a/testing/test_convert_nodes.py
+++ b/testing/test_convert_nodes.py
@@ -1,15 +1,148 @@
-"""The Blender 5.2+ identity-weld convert node group.
-
-Only runs on 5.2+, where the ``Merge Points`` node exists. On older Blender the
-convert modifier keeps loading the merge-by-distance asset, so there is nothing
-to build here. Besides weld identity, the 5.2 group materializes deterministic
-vertex/face ids on the generated mesh for consumers that keep element links.
-"""
+"""Stable generated ids for both CAD Sketcher conversion paths."""
 
 import unittest
 from unittest import TestCase
 
 import bpy
+
+
+class TestGeneratedIds(TestCase):
+    def test_active_conversion_path_emits_stable_ids(self):
+        from ..utilities.convert_nodes import (
+            FACE_ID_ATTR,
+            VERTEX_ID_ATTR,
+        )
+        from ..utilities.curve_data import _get_convert_node_group
+
+        group = _get_convert_node_group()
+        self.assertIsNotNone(group)
+        stores = {
+            node.inputs["Name"].default_value: node
+            for node in group.nodes
+            if node.bl_idname == "GeometryNodeStoreNamedAttribute"
+        }
+        self.assertEqual(stores[VERTEX_ID_ATTR].domain, "POINT")
+        self.assertEqual(stores[FACE_ID_ATTR].domain, "FACE")
+        self.assertFalse(
+            any(
+                node.bl_idname == "GeometryNodeInputIndex"
+                for node in group.nodes
+            ),
+            "generated ids must not depend on topology-global Index",
+        )
+
+    def test_unaffected_ids_survive_neighbor_insert_and_remove(self):
+        """Re-evaluate after global reindexing and compare unaffected elements."""
+        from ..utilities.convert_nodes import (
+            FACE_ID_ATTR,
+            SOURCE_CURVE_ID_ATTR,
+            SOURCE_ENDPOINT_ID_ATTR,
+            VERTEX_ID_ATTR,
+            add_generated_id_nodes,
+        )
+
+        group = bpy.data.node_groups.new(
+            "test_stable_generated_ids", "GeometryNodeTree"
+        )
+        group.interface.new_socket(
+            "Geometry", in_out="OUTPUT", socket_type="NodeSocketGeometry"
+        )
+        group.interface.new_socket(
+            "Geometry", in_out="INPUT", socket_type="NodeSocketGeometry"
+        )
+        nodes, links = group.nodes, group.links
+        group_input = nodes.new("NodeGroupInput")
+        group_output = nodes.new("NodeGroupOutput")
+        geometry = add_generated_id_nodes(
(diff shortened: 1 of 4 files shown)


MyTH-zyxeon:
Thanks for tackling the stable-identity bounty. I reviewed exact head `57a5e1d46746aeb79cd3a9f689c1b838737c7c31` against issue #6 and found two acceptance blockers behind the green checks:

1. **The feature is absent on the minimum supported Blender release.** `blender_manifest.toml` declares Blender 5.0 as the lower bound, and this head's CI reports a green `test-extension (blender 5.0)` job. However, `utilities/convert_nodes.py` explicitly builds the new attribute stores only for Blender 5.2+, while older Blender continues loading the shipped merge-by-distance asset. `testing/test_convert_nodes.py` is likewise 5.2+-only, so the 5.0 job cannot prove or expose the missing identifiers. A 5.0 user still receives no `cad_sketcher_vertex_id` / `cad_sketcher_face_id` output.

2. **Raw generated indices do not preserve element identity through topology edits.** Both attributes are populated from `GeometryNodeInputIndex`. Those values are repeatable only while the generated topology and ordering remain unchanged, as the new comments and test acknowledge. Inserting or deleting geometry before an existing vertex/face reindexes the output, so MeasureIt-style links to unaffected existing elements can silently move to different elements. The regression test verifies only node wiring and has no before/after topology-edit assertion.

To close #6, the conversion needs IDs derived from stable source geometry identity (with a deterministic child identity for tessellated/fill output), carried through both supported conversion paths, plus a regression that records unaffected IDs, inserts/removes neighboring topology, reevaluates, and proves those IDs still select the same elements on Blender 5.0 and latest.

Run-Id: `run-20260815-0b0cd1a7-1945-40de-85f1-dfef5094351b`
Trace-Id: `61c8fdc8-ae16-4fc2-8a82-badc6ef1e4d5`
Requester: Jun local owner via scheduled automation `monitor-sxt-bounty-pr-1757`
Implementer: MISA 3 bot ID `1516725819517567077`

franklincg:
Addressed both acceptance blockers from the previous review on current head `0dd2d9c4cfcc335433524e77d01136bce6244fcd`:

1. The generated identity path is now carried through the supported conversion paths, including Blender 5.0 coverage.
2. Generated POINT/FACE IDs are derived from stable source identity plus source-local child identity rather than raw topology-global Index. Regression coverage now records unaffected IDs, inserts/removes neighboring topology, re-evaluates, and verifies the unaffected elements retain their identities.

I also re-checked the current head against the two requested acceptance conditions, and the current GitHub Actions runs are green for lint, test, and perf.

Could you please re-review the updated head?

hlorus:
Thanks @franklincg, cleanly built and the UUID-derived, source-local approach is the right foundation.

The thing to align on is who consumes these ids. Index-based addons like MeasureIt (the use case behind #6) link by vertex/face index and will not read a custom `cad_sketcher_vertex_id`, so as-is this does not make those tools work. That is inherent, index references cannot survive topology edits, stable linking requires the consumer to key off a persistent id. So this really delivers a stable-id primitive, valuable only once something reads it.

Given that, we would rather write the vertex id into Blender's reserved `id` attribute and document `id` as the public link key (faces still need a custom attribute, there is no standard face-domain `id`), and verify it survives to the evaluated mesh.

What consumer or use case did you have in mind for these ids?

franklincg:
MeasureIt is the motivating use case I had in mind, based on the original #6 discussion. I agree that its current index-based references will not survive topology edits by themselves.

The intent here is to provide the persistent identity primitive on the evaluated geometry: vertices through Blender's reserved id attribute and faces through the custom face-domain attribute. A consumer can then key references by those persistent IDs instead of transient mesh indices.

So I'm not claiming this makes the existing MeasureIt index behavior persistent without consumer-side changes; it establishes the stable key needed for that integration.

hlorus:
Merged, thanks @franklincg. Publishing the stable vertex identity through Blender's reserved `id` attribute (with faces on a custom attribute) is exactly the right shape, and the evaluated-mesh and neighbor-edit tests give it solid coverage.

This closes a bounty issue (#6, Bounty - Small), so you are welcome to collect it: submit an invoice on our Open Collective at https://opencollective.com/cadsketcher/expenses/new, choose Invoice, and reference `Keep generated mesh element identifiers stable through edits #611`. Thanks again for contributing!

franklincg:
Hi hlorus,

Quick question regarding the approved CAD Sketcher expenses on Open Collective.

A few of them are still showing as Pending after the earlier PayPal/Wise payout issues were resolved. I just wanted to check whether there is anything still needed from the CAD Sketcher side, or if they are now simply waiting on Open Collective's payment processing.

No rush — just making sure I’m not missing any action on my side.

Thanks!

franklincg:
Hi @hlorus, quick follow-up on the three Open Collective expenses I mentioned earlier.
Open Source Collective confirmed today that #342673, #342262 and #342109 are currently pending collective approval again, and they need a collective admin to re-approve them before payment can be processed.

Could you please re-approve those three expenses when you have a moment?

Thanks!

hlorus:
Ok, i've re-approved the three expenses, let's hope it works this time!

PR #612, opened 2026-08-14
Preserve custom attributes through native conversion (#200)

### LINKED ISSUE #200: Allow setting properties/attributes on converted elements
When convertion happens already converted geometry gets overwritten, this can lead to a loss of data when the user manually modifies anything in a destructive manner on the resulted object data and reruns the conversion. #104 is an example of this limitation.

To allow changing properties and attributes there needs to be some way to set these values without them getting overwritten.

The attributes branch tries to solve that by implementing a system that writes addon defined attributes whenever the conversion happens. 

Note: Conversion always involves a bezier curve which don't yet support geometry attributes, therefor attributes are currently limited to have per object/object data values

Design:
- Adds a CustomAttribute property group which stores the name, type and domain of the attribute definition
- Attribute definitions are stored per sketch
- Attribute values are stored per entity in an id property e.g. entity["attribute_values"] = {"some_integer": 12}
- Values can be set on sketches, this acts as a default if no value is set on an entity
Conversion :
- Attributes of domain type "OBJECT" will try to set the value on the converted object and object data
- Attributes of other domain types will set the property with the matching name on the mesh elements or set an geometry attribue with the same settings 

Alternatives:
Another approach would be to try to reuse existing geometry when converting.

franklincg:
Fixes #200.

Adds user-defined attributes to the current native-Curves source-of-truth architecture.

- stores attribute definitions on the sketch Curves datablock;
- supports BOOLEAN / INT / FLOAT values on POINT, segment (CURVE), and OBJECT domains;
- sketch-level defaults are applied to geometry that already exists and to native curves created later;
- values are authored on native sketch geometry instead of disposable converted output;
- custom named attributes survive native curve rebuilds and remain available to the conversion source path;
- adds Add / Set / Remove operators and an Attributes panel;
- regression coverage for segment, point and object values, future-geometry defaults, refresh persistence, removal and operator registration;
- fork CI passes lint, Blender 5.0, Blender latest, and perf.

Bounty: Small ($100).


MyTH-zyxeon:
I reviewed exact head `964362bd04399c44115397470ce5575b49d64253` against the acceptance design in #200 and found three gaps:

1. **`OBJECT` attributes never reach object data.** `define_attribute()` and `set_attribute_value()` only write `sketch.target_object[name]` (`utilities/custom_attributes.py:109-111`, `154-156`). `Sketch.target_object` is the Curves object, while `Sketch.data` is its datablock (`model/sketch_ref.py:27-37`). The issue explicitly requires trying to set `OBJECT` values on both the converted object and object data, so the datablock half is currently missing (and removal only clears the object property too).

2. **Opening Set can silently replace an existing/default value with zero.** The Set operator's typed fields always start as `False` / `0` / `0.0` (`operators/custom_attributes.py:85-88`), and `invoke()` only resolves the definition; it never initializes the active field from the selected entity's current value or the definition default (`94-99`). Confirming the dialog without re-entering a value therefore clobbers non-zero values. Please seed the field from the common selected value (falling back to the definition default) and cover that behavior.

3. **The conversion acceptance boundary is still untested.** `test_point_attribute_survives_conversion_refresh` explicitly stops at the native Curves source because the background runner does not materialize the evaluated Mesh (`testing/test_custom_attributes.py:85-106`). That proves source persistence, but #200 is specifically about values surviving onto the converted object/data and generated elements. Please add a regression that inspects the evaluated/conversion result (or refactor the propagation step behind a directly testable boundary) and asserts OBJECT plus POINT/CURVE values on the produced output.

Run-Id: `run-b73e515d-fa4d-4e3b-9460-e395b9101539`
Trace-Id: `eb131fb4-44a4-480b-ad37-66d5bf784a5d`
Requester: Codex heartbeat automation `monitor-sxt-bounty-pr-1757`
Implementer: MISA 3 bot ID `1516725819517567077`

franklincg:
Thanks for the detailed review. I addressed all three gaps on the current head `310fdb2ec93ede4ca5658a6210d8e73560097059`:

1. OBJECT custom attributes are now mirrored to both the sketch object and its datablock, including update/removal paths.

2. The Set dialog now seeds from the common selected value, with the attribute default used for mixed/empty selections, so confirming the dialog no longer silently resets values to zero.

3. Added a real evaluated conversion regression (`test_named_attributes_reach_evaluated_conversion_output`) that exercises the conversion modifier and verifies POINT/CURVE named attributes on the produced output, plus OBJECT values on both object and object data.

Current CI is green:
- lint ✅
- tests ✅
- perf ✅

Could you please re-review the updated head?

hlorus:
Thanks @franklincg, the native-attributes direction is right.

Main question: most of the complexity here (per-sketch groups rebuilt on change, explicit re-store) is only to carry attributes through conversion, but the values already live on the source curve and GN propagates named attributes generically. So the loss is at a specific node in the convert chain, likely the mesh-to-curve -> fill -> mesh round-trip or the weld. Did you try relying on propagation, and where did it break? If it is a fixable/avoidable step, the convert group could transport attributes generically as a single shared group, removing most of this.

Also: this now conflicts with main. #611 (stable ids) just reworked `convert_nodes.py`, the same file. Needs a rebase and hand-merge, and add `id` to `_RESERVED_NAMES` while there.

franklincg:
Hi @hlorus,

I’ve finished the rework based on your feedback and resynced the branch with the current main.

The implementation now:
- keeps user-defined values on native Curves source attributes;
- reserves Blender’s `id` attribute;
- preserves OBJECT values on both the object and datablock;
- transfers POINT/CURVE custom attributes through the native conversion path;
- avoids zeroing existing/default values in the Set flow;
- handles Blender 5.2 destructive conversion replacing the Curves datablock.

Current CI is fully green:
- lint ✅
- Blender 5.0 ✅
- Blender latest ✅
- perf ✅

The PR is mergeable and ready for re-review.

Thanks!

hlorus:
Thanks @franklincg. I did some testing on whether GN can carry the attributes generically, and I think this can be a lot simpler than the current schema-variant + Sample-Nearest approach.

Findings (Blender 5.2, isolated tests):

- Curve to Mesh, Merge, and Mesh to Curve all propagate named POINT/CURVE attributes generically, no extra nodes. Fill Curve is the only node that drops them.
- So the wire/extrude path needs zero attribute machinery, and only the fill needs a bridge.
- Sample Nearest onto the POINT domain is unreliable for per-segment values: at a shared corner the Merge step averages the two segments' values. With four edges tagged 10/20/30/40 the corners came out 15/25/35, which is meaningless for something like material_index.
- Capturing the per-segment value on the EDGE domain (a single Store Named Attribute before the merge) gives the exact values 10/20/30/40 through the weld, deterministically, no spatial matching.

So the reliable recipe is domain-correct capture rather than Sample Nearest: POINT attributes propagate as-is, per-segment values go on EDGE (maps cleanly to extrude walls), and a per-region value is one value on the closed curve for fill faces. A flat fill's interior faces have no single owning segment, so that case is ambiguous by definition, not something a sampler can fix.

Given that, I think we should drop the per-schema variant groups and the Sample-Nearest transfer and go with the generic + domain-capture approach. It is both simpler and more correct. I have a small prototype confirming the above end to end and can share it if useful.

franklincg:
Thanks @hlorus — I reworked the conversion path around your domain-capture findings.

- Removed the per-schema converter variants.
- Removed Sample Nearest / spatial transfer entirely.
- Kept one shared `CAD Sketcher Convert` group.
- POINT attributes follow the generic GN propagation path.
- Per-segment CURVE values are re-homed onto EDGE before Merge Points.
- The non-fill wire path keeps the welded edge representation, so segment values such as 10/20/30/40 survive exactly instead of becoming 15/25/35 at shared corners.
- Fill uses a domain-aware bridge without spatial matching. Flat fill faces are not assigned a per-segment ownership value because there is no unique owning segment.
- OBJECT values continue to survive destructive conversion on both the object and object data.

I added regression coverage specifically for the 10/20/30/40 case and the averaged 15/25/35 values are explicitly rejected.

Current head: d026c44f

All CI is green:
- lint ✅
- tests ✅
- perf ✅

Could you please retest/re-review this head when you have a chance?

hlorus:
Thanks for pushing this forward. The Attributes panel (add / set / remove, with the domain shown per entry) is clean and exactly the right UX, that part looks good to me.

My feedback is on the backend, which feels heavier than the problem needs. The conversion side is already in good shape: letting named POINT/CURVE attributes propagate and only bridging the two lossy spots (the EDGE re-home before the weld, and the capture/restore across Fill Curve) is the right call, and it avoids the corner ambiguity that spatial sampling had.

Where I think it can slim down is the plumbing around that:

- The object-level value mirror (`OBJECT_NAMES_PROP`) plus the `sync_object_attribute_values` depsgraph handler is a lot of surface just to keep a second copy of values in step. If a custom attribute is simply a native attribute on the Curves data, there is one source of truth and no per-update sync is needed.
- Definitions live in two places (the curve-data defs and the object names list) and are aggregated globally via `_sync_shared_conversion_group` to feed the one shared convert group. That global coupling is the part that gets hardest to reason about.

I opened #630 as a small draft that isolates just the conversion mechanism (the same domain-correct capture, minus the storage/sync layer) with a test that a 10/20/30/40 loop survives the weld and the fill exactly. I think the ideal shape is: keep this UI, treat attributes as plain native Curves attributes, and lean on that minimal conversion path, dropping the object-value sync (or deferring object-domain attributes to a follow-up).

Happy to help wire the panel onto the thinner backend if that is useful.

franklincg:
Thanks for the guidance. I updated the PR to follow the thinner backend direction:

- kept the Attributes UI;
- removed the object-level value mirror and depsgraph sync;
- deferred OBJECT-domain attributes;
- kept POINT/CURVE values as native Curves attributes;
- aligned the conversion path with the approach demonstrated in #630;
- added regression coverage for the 10/20/30/40 segment values surviving weld + Fill without corner averaging.

Lint, Blender tests, and perf are all green now.

Could you please take another look when you have a chance?

hlorus:
Found a regression in the conversion path while testing: defining a segment attribute turns Fill off on existing sketches.

Repro: draw a filled rectangle (closed loop), then define a CURVE-domain attribute on it. The fill disappears (faces drop to zero).

It is not the attribute node technique. The EDGE re-home and the Fill bridge are fine; bypassing both does not bring the fill back, and the mesh is still welded correctly (same vertex count before and after, `merge_id` intact). The output after adding the attribute is exactly the welded wire, which is the Switch node's non-fill branch.

The cause is in `build_convert_node_group`: when the attribute signature changes it rebuilds with `ng.interface.clear()` and recreates the sockets. That re-mints the socket identifiers (the Fill input went from `Socket_2` to `Socket_5` in my run). Every already-bound sketch modifier stored its Fill value under the old identifier, so after the rebuild that value is orphaned, and an unset boolean modifier input evaluates as False. The Switch then outputs the wire instead of the fill. So defining an attribute silently flips Fill off on every existing sketch.

Suggested fix: do not clear and recreate the interface on rebuild. The bridge is purely internal named-attribute nodes and needs no new interface sockets, so keep the three sockets (Geometry in, Geometry out, Fill) stable and only rebuild the internal nodes and links. That preserves modifier bindings and their Fill values. A fallback would be re-initializing Fill on bound modifiers after a rebuild, but keeping the interface stable is cleaner.

Worth a fill-survival check in the attribute test too, a value-only assertion passes right through this.

Happy to push the interface-preserving change to the shared converter if useful.

franklincg:
Hi @hlorus,

I addressed the Fill regression you found.

The shared `CAD Sketcher Convert` group now preserves its existing public interface when the attribute schema changes. It rebuilds only the internal nodes/links, so the `Fill` socket identifier is no longer reminted and existing modifier bindings remain intact.

I added regression coverage that verifies:

- the Fill socket identifier is unchanged across an attribute-schema rebuild;
- the existing Fill value remains enabled;
- defining a CURVE-domain attribute on an already-filled sketch keeps the converted result filled (faces remain present);
- per-segment values on the wire/EDGE path still preserve the exact 10/20/30/40 values without the 15/25/35 corner averaging.

As discussed earlier, flat-fill interior faces are not assigned arbitrary per-segment ownership, and OBJECT-domain attributes remain deferred.

I also removed the temporary diagnostic tests used while isolating the Blender 5.2 behavior.

Current head:
2bf8688419e114b4d717d56439e7d9c5b0e54d4a

Current CI:
- lint ✅
- Blender 5.0 ✅
- Blender 5.2/latest ✅
- perf ✅

The PR is mergeable and ready for another review.

Thanks!

hlorus:
Thanks @franklincg! This feature has landed in `main` — merged with additional changes in #639, where you're credited as co-author.

This closes a bounty issue (#200, Bounty - Small), so you are welcome to collect it: submit an invoice on our Open Collective at https://opencollective.com/cadsketcher/expenses/new, choose Invoice, and reference `Allow setting properties/attributes on converted elements #612`. Thanks again for contributing!

franklincg:
Hi @hlorus — just a quick follow-up on the Small bounty for #200 / #612.
I submitted the Open Collective invoice you invited me to create (expense #343710) on Aug 25, and it is still showing as pending on my side.
When you have a chance, could you please take a look and approve/release it if everything is in order?
Thanks again!

hlorus:
Sorry, missed that one. It's approved now!

PR #682, opened 2026-09-08
Add Create Mesh Output: live, exportable mesh mirror of a sketch

hlorus:
A **Create Mesh Output** tool that spawns a companion **mesh** object mirroring
the active sketch live — giving a real, exportable, `to_mesh`-able mesh that a
Curves-typed sketch object can't provide on its own.

## What

- `CAD Sketcher To Mesh` group: `Object Info(sketch, Relative) → Realize → Output`.
- The object's transform is matched to the sketch, so the pulled geometry is
  planar in local space yet lands in the correct world position.
- The sketch is bound through the modifier's Object input (version-aware
  `set_modifier_input`).

Result: a live, exportable mesh that follows the sketch, respects its Fill/wire
setting, carries the generated vertex/face ids through, and can be **Applied** to
bake a static editable mesh.

## Integrates with managed collections (#685)

Rebased on the merged collection work. The companion is grouped into its **source
sketch's collection** via `place_with_sketch`, so the consumable stays with its
part (falls back to the scene collection for a legacy/unmanaged sketch).

## Tests

- `test_mesh_output_is_a_readable_ring` — companion is a `MESH`, `to_mesh()`
  works, carries the correct ring.
- `test_mesh_output_lands_in_the_sketch_collection` — grouped with its sketch,
  not loose in the scene root.
- `test_mesh_output_follows_a_transformed_sketch` — correct world position on a
  non-XY workplane.
- Full suite: `Ran 456 tests ... OK` (2 pre-existing expected failures).

## Follow-ups (not in this PR)

- Rename the mesh when the sketch is renamed.
- Deleting the sketch leaves the modifier's Object input empty (empty mesh, no
  crash) — no active cleanup.

### DIFF
--- a/declarations.py
+++ b/declarations.py
@@ -65,6 +65,7 @@ class Operators(str, Enum):
     DeleteConstraint = "view3d.slvs_delete_constraint"
     DeleteEntity = "view3d.slvs_delete_entity"
     DeleteSketch = "view3d.slvs_delete_sketch"
+    CreateMeshOutput = "view3d.slvs_create_mesh_output"
     MergePoints = "view3d.slvs_merge_points"
     Paste = "view3d.slvs_paste"
     Move = "view3d.slvs_move"
--- a/operators/__init__.py
+++ b/operators/__init__.py
@@ -26,6 +26,7 @@ modules = [
     "bevel",
     "offset",
     "project_geometry",
+    "mesh_output",
     "set_sketch",
     "delete_entity",
     "delete_sketch",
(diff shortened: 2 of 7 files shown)


hlorus:
Superseded by #725, which is merged: a sketch's geometry is now always realised on a mesh body rather than an optional companion mesh, so the result applies, exports and edits like any other mesh (the problem this PR set out to solve, and #723). Thanks for the groundwork here -- the Object Info approach and the node group came from it.

PR #694, opened 2026-09-12
Interaction performance on large sketches

hlorus:
Interaction performance on large sketches, prompted by a Discord report that
"you need a beefy computer to run CAD Sketcher". Seven changes, each measured.
No behaviour is meant to change; it is the same geometry computed less often and
with less Python in the way.

## Result

One drag frame (solve + mesh refresh + overlay + picking), 200 circles:

| | before | after |
|---|---|---|
| drag frame | 44.0 ms (23 fps) | 9.9 ms (101 fps) |
| hover frame (no geometry change) | 14.9 ms | 0.28 ms |
| per-mouse-move segment distance test | 18.5 ms | 0.66 ms |
| 400 circles, drag | ~90 ms (11 fps) | 22.0 ms (45 fps) |
| 200 distance constraints, solve | 29.0 ms | 23.2 ms |

And the one that is not about sketch size at all: **solve time used to scale with
how many unrelated objects the file contained.** Identical sketch, 100
constraints, only the surrounding scene changing:

| scene objects | before | after |
|---|---|---|
| 4 | 10.3 ms | ~9 ms |
| 204 | 24.5 ms | ~9 ms |
| 504 | 47.6 ms | ~9 ms |
| 1004 | 81.1 ms | ~9 ms |

If the report came from someone with a real scene rather than a big sketch, that
cliff is the likeliest explanation.

## The changes

**1. Tessellate all arcs/circles in one vectorized pass.** `_arc_points` built
each arc in a Python loop doing one `mat @ Vector(...)` per point, 58% of the
extraction on a circle-heavy sketch. Worth knowing for anyone revisiting it:
vectorizing a *single* arc is slower, because 49 points is too few to amortize
the numpy setup. Only batching across all curves wins.

**2. Solve rebuilds only the segments whose points actually moved.**
`_write_results` called `rebuild_segments(sketch)` unfiltered, re-deriving every
arc/circle bezier on every solve, even though the `point_ids` filter already
existed and `batch_update` already used it. Positions are compared at float32,
the attribute's own storage, so "moved" means "the stored bytes change" with no
tolerance to tune.

**3. One geometry extraction shared by the overlay and the picker.** Both kept
their own cache and ran their own extraction, so every changed frame paid twice.
They keyed on different signatures (picking on geometry, the overlay on geometry
plus selection), so this is a split rather than a shared cache: a
selection-independent geometry pass cached on `geometry_signature`, and a cheap
colour pass. Hover now redoes only colours.

**4. Geometry stays in numpy end to end.** Segment identity moved from
per-segment to per-curve (`seg_cids` plus a `seg_owner` row map), so a circle
carries one curve_id instead of 48 — that bookkeeping was what forced everything
into Python lists. `batch_for_shader` and the screen projection both take arrays
directly.

This also fixed something I had under-counted: `pick_ranked` runs on every
mouse-move and measured point-to-segment distance in a Python loop over every
tessellated segment (18.5 ms at 200 circles, now 0.66 ms, 28x).

**5. Skip the merge-id recompute when connectivity is unchanged.**
`compute_merge_ids` and `compute_generated_id_seeds` derive weld ids purely from
connectivity: which point curves a segment references, curve types, per-curve
point counts, and coincidence constraints. Never from positions. They ran on
every refresh and every solve anyway. Now gated on a connectivity signature,
with `force=True` to override. `refresh_curve_geometry` 4.0 → 1.2 ms.

**6. Resolve a constraint's sketch without scanning the scene.** This is the
scene-size cliff above. `GenericConstraint._get_sketch` found the owning Object
by scanning `bpy.data.objects`, ~7 times per constraint per solve (700 calls for
100 constraints), so the cost was O(constraints x objects), paid every drag frame.

A gotcha worth recording: my first attempt cached the owner's *name* and looked it
up with `bpy.data.objects.get()`. That was only 2.2x better, because
**`bpy.data.objects.get()` is itself a linear scan**, not a hash lookup — 0.77 us
at 10 objects, 48.8 us at 2000. The cache holds the Object reference instead.

**7. Publish the solved sketch so constraints stop re-deriving it.** Even cached,
`_get_sketch` ran ~2800 times per solve on a 400-constraint sketch. The solver
already knows which sketch it is solving, so it publishes it for the duration of
the system build. 20% off a constrained solve.

I deliberately did *not* thread a parameter through all 13 constraint types for
this. The override is honoured only when the constraint's own `id_data` IS that
sketch's data, so a constraint from anywhere else falls through to the normal
lookup and cannot resolve to the wrong sketch. That guard is what made the narrow
change safe.

## Behaviour changes to be aware of

- A circle now emits exactly 48 segments instead of 49. The old code closed the
  loop with a duplicated final point and a zero-length segment, which was also a
  wasted pick candidate.
- `SketchRenderData.line_buckets` / `point_buckets` now hold numpy arrays, and
  `point_ids` / `segment_ids` are replaced by arrays on `SketchGeometry`.
  `render_data.geometry(sketch)` is the entry point for pickable geometry.
- `picking.rank_hits` takes a `PickSet` instead of two lists of tuples.
- `compute_merge_ids(sketch, force=True)` forces the old unconditional behaviour.

## What to hammer when testing

Selection and hover are what changes 1-4 touch most: hovering and clicking
overlapping entities (the #50 cycling path), box select, construction geometry
colours, selected/hovered point sizes, arcs growing and shrinking past their
sweep boundaries, dimension values driving geometry, and reference-geometry
picking across several curve objects. Also the Vulkan backend, since the
constant-draw-call batch structure is what this rests on.

For change 5, **fill and boolean/extrude output** specifically, since those
consume the weld ids.

For changes 6 and 7, **undo/redo on a constrained sketch**, and sketches in a
file with many objects. See the gap below.

## Testing done

496 tests pass (38 new), 1 skipped, 2 pre-existing expected failures, run against
this branch built and installed as `bl_ext.user_default.CAD_Sketcher` in an
isolated `BLENDER_USER_RESOURCES`. All CI green.

New coverage is concentrated where a mistake would be *silent* rather than loud:

- `test_solve_scoped_rebuild.py` runs an unscoped rebuild after each solve and
  asserts nothing changes, across point drags, arc/circle centre moves,
  diameter-driven resizes, arcs resegmenting, solves after topology changes,
  coincident-corner weld ids, and settled no-op solves. This net caught two real
  gaps while I was writing it (the radius case, and merge ids).
- `test_merge_id_gating.py` performs an edit, lets the gate decide, then forces a
  recompute and asserts the weld ids and source seeds are byte-identical.
  Including the case that needs the constraint in the signature: two endpoints at
  the same spot are distinct point curves until a coincidence constraint unions
  them, so that constraint changes the answer with every position, id and count
  unchanged.
- `test_constraint_sketch_resolution.py` covers a deliberately **poisoned** cache
  entry losing to a rescan, a cached reference to a deleted object, two sketches
  resolving separately under interleaved access, the override being refused for a
  foreign constraint in both directions, the override not leaking out of a raised
  exception, and two sketches with different distance values solving end to end.
- `TestArcTessellation` guards the tessellated geometry (every vertex on-radius,
  circle chain closed with no gap, no zero-length segments, arcs landing exactly
  on their defining points at 30/90/200/350 degrees).
- `TestSegmentDistance` pins the vectorized distance against the scalar formula it
  replaced, including degenerate and clamped cases.
- Tests that one frame extracts once and a hover extracts zero times, so a
  regression to double extraction fails the suite rather than quietly costing
  performance. `scripts/perf_harness.py` gained the same as deterministic
  op-count metrics.

The test suite has no GPU, so I separately built the real `_build_batches` output
from real sketch data on a live `gpu.init()` context across five selection states
and confirmed all three batches construct with the draw-call count staying
constant.

### Known gap

The undo/redo test **skips** headless (`bpy.ops.ed.undo` needs a real context), so
undo is the one case with no automated coverage — and it is the classic way a
cached `bpy` reference breaks. I expect it to be fine: invalidated references
raise `ReferenceError`, which is caught, and a cached Object is re-verified to
still own the datablock on every use, so a stale entry costs a rescan rather than
returning the wrong object. But I cannot prove it headless, so please exercise it.

## Not addressed

For a constrained sketch, a good deal of the solve is still Python building the
system: the `ref()` -> `curve_ref` -> `get_curve_type` -> `get_curve_data` chain
runs thousands of times per solve, plus ~8400 function-local import resolutions.

I can measure that stubbing out `_init_constraints` drops a 200-constraint solve
from 23.2 ms to 3.3 ms, but that is *not* a fair attribution: removing the
constraints also leaves slvs an empty system, so it mixes Python overhead with
real solver work. The honest next step is timing instrumentation inside
`_solve_once` to get the true split, because if slvs is most of the remainder then
further Python work buys little.

Measurements are from one machine (Blender 5.2, Linux); treat them as trends, not
absolutes.

### DIFF
--- a/blender_manifest.toml
+++ b/blender_manifest.toml
@@ -1,7 +1,7 @@
 schema_version = "1.0.0"
 
 id = "CAD_Sketcher"
-version = "0.31.2"
+version = "0.32.0"
 name = "CAD Sketcher"
 tagline = "Parametric, constraint-based geometry sketcher"
 maintainer = "hlorus <dave19924@gmail.com>"
--- a/curve_solver.py
+++ b/curve_solver.py
@@ -11,6 +11,7 @@ Constraints still come from PropertyGroups (referencing entities for now).
 import logging
 import math
 
+import numpy as np
 from mathutils import Matrix as _Matrix
 from mathutils import Vector
 
@@ -24,6 +25,19 @@ from .utilities.curve_data import (
 from .utilities.workplane import ensure_workplane_empty
 
 logger = logging.getLogger(__name__)
+
+
+def _stored_differs(current, new) -> bool:
+    """Whether writing ``new`` over ``current`` would change the stored floats.
+
+    Positions are float32 attributes, so the comparison is done at that precision
+    rather than against the solver's doubles (see ``_write_results``).
+    """
+    return bool(
+        (
+            np.asarray(new, dtype=np.float32) != np.asarray(current, dtype=np.float32)
+        ).any()
+    )
 
 
 class CurveSolver:
@@ -456,7 +470,20 @@ class CurveSolver:
         cid_list = read_uuid_list(curve_data, "curve_id")
         cp_list = read_uuid_list(curve_data, "center_point_id")
 
+        # Point curves whose solved position actually differs from what is stored.
+        # The third pass rebuilds only the segments referencing these, instead of
+        # re-deriving every arc/circle bezier in the sketch on every solve -- a
+        # drag moves a handful of points out of hundreds (issue #342).
+        #
+        # A segment's geometry is a pure function of its referenced point
+        # positions, so a point the solver rewrites with an unchanged value cannot
+        # change any segment and is deliberately left out. That relies on the
+        # invariant every writer of point positions already maintains: whoever
+        # moves a point rebuilds its segments (see batch_update).
+        moved_point_ids = set()
+
         # First pass: update all point positions
+        solved_indices, solved_positions, solved_cids = [], [], []
         for curve_idx in range(n_curves):
             ctype = type_attr.data[curve_idx].value
             cid = cid_list[curve_idx]
@@ -464,8 +491,23 @@ class CurveSolver:
             if ctype == SketchCurveType.POINT:
                 pos = self._get_solved_point_position(cid)
                 if pos:
-                    pt_idx = curve_data.curves[curve_idx].points[0].index
-                    curve_data.points[pt_idx].position = pos
+                    solved_indices.append(curve_data.curves[curve_idx].points[0].index)
+                    solved_positions.append(pos)
+                    solved_cids.append(cid)
+
+        if solved_indices:
+            # Compare as float32, the attribute's own storage: the solver works in
+            # double, so a solved value that merely round-trips would never equal
+            # the stored one under an exact double compare, and every point would
+            # look moved. Comparing what would actually be *stored* makes "moved"
+            # mean "the bytes change", with no tolerance to tune.
+            stored = np.empty(len(curve_data.points) * 3, dtype=np.float32)
+            curve_data.points.foreach_get("position", stored)
+            current = stored.reshape(-1, 3)[solved_indices]
+            incoming = np.asarray(solved_positions, dtype=np.float32)
+            for k in np.flatnonzero((incoming != current).any(axis=1)).tolist():
+                curve_data.points[solved_indices[k]].position = solved_positions[k]
+                moved_point_ids.add(solved_cids[k])
 
         # Second pass: update circle/arc edge positions from solved radius
         for curve_idx in range(n_curves):
@@ -485,11 +527,19 @@ class CurveSolver:
                         # Update first edge point at new radius
                         curve_slice = curve_data.curves[curve_idx]
                         first = curve_slice.points[0].index
-                        curve_data.points[first].position = (
+                        edge = (
                             ct_pos[0] + solved_radius,
                             ct_pos[1],
                             ct_pos[2],
                         )
+                        point = curve_data.points[first]
+                        if _stored_differs(point.position, edge):
+                            point.position = edge
+                            # This circle's radius changed but its centre may not
+                            # have moved, so nothing above would have flagged it.
+                            # rebuild_segments matches a circle through the centre
+                            # point it references, so that is the id to add.
+                            moved_point_ids.add(cp_id)
 
         # Third pass: rebuild segments from updated point positions.
         if getattr(sketch, "is_3d", False):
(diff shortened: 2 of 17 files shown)


hlorus:
Added a fifth change: **skip the merge-id recompute when connectivity is unchanged** (`080831e9`).

`compute_merge_ids` (and `compute_generated_id_seeds`, which it calls) derive weld ids purely from connectivity: which point curves a segment references, curve types, per-curve point counts, and coincidence constraints. Never from positions. But they ran on every `refresh_curve_geometry` and every solve, so a drag re-derived them dozens of times a second for an answer that could not have changed.

They are now gated on a connectivity signature, with `force=True` to override.

Drag frame at 200 circles:

| | before this commit | after |
|---|---|---|
| `refresh_curve_geometry` | 3.97 ms | 1.17 ms |
| `solve_system` | 7.21 ms | 5.80 ms |
| **drag frame** | 14.2 ms (70 fps) | **10.0 ms (100 fps)** |
| 400 circles | 28.1 ms (36 fps) | 20.0 ms (50 fps) |

Solve improved too, because the explicit `compute_merge_ids` call I added there for the scoped rebuild now skips as well. What remains in `refresh_curve_geometry` is its own attribute save/restore, which is the actual required work; the merge pass and the signature are about 0.1 ms each.

### Safety

Stale weld ids do not raise, they stop a corner welding and the sketch quietly fails to fill, so `testing/test_merge_id_gating.py` performs an edit, lets the gate decide, then forces a recompute and asserts `merge_id` and both source-seed attributes are byte-identical. Covered: position-only changes, adding a segment, adding a coincidence constraint, arc resegmentation, after a geometry refresh, and a datablock whose outputs were deleted (the gate must never skip when the attributes do not exist yet).

The coincidence case is the one worth flagging: two endpoints at the same spot are distinct point curves until a constraint unions them, so that constraint changes the answer with every position, id and count unchanged. It is in the signature, and a test asserts the signature moves when it is added.

There is also a separate `TestConnectivitySignature` that tests the signature's discrimination directly. Asserting on `compute_merge_ids`' return value cannot show this, because creating a curve recomputes the ids on the way through, so by the time a test looks the signature is already current again. That tripped up my first attempt.

481 → 484 tests pass (11 new), 2 pre-existing expected failures. `test_fill_coincident` (the real fill path) passes.

### Session total, 200 circles

Drag frame 44.0 ms → 10.0 ms, 23 → 100 fps. Hover frame 14.9 ms → 0.29 ms, plus the per-mouse-move segment distance test that went from 18.5 ms to 0.66 ms.

Worth re-testing fill and boolean/extrude output specifically, since those consume the weld ids.

